Issue #24, opened 2023-04-12
Support serving the dashboard from a non-root path (ngnix proxy)


swastis10:
While serving the dashboard from a non-root path - e.g. using an ngnix reverse proxy, we need to special configuration. Otherwise, dash tries to load the assets from the root URL, which fails.

We use an ngnix reverse proxy in the NREL hosted environment, where the app is hosted at
https://[deployment]-openpath.nrel.gov/admin

The application was working fine when we are testing on localhost because the app is served from the root URL but when we deploy, the app is served from the /admin URL.

The error we get is:
<img width="365" alt="image" src="https://user-images.githubusercontent.com/124807088/231607429-8e971e52-2da3-4fde-91cc-9af6bc7b4317.png">

In order to fix it, we have used [url_base_pathname](https://dash.plotly.com/reference)
>> A local URL prefix to use app-wide. Default '/'. Both requests_pathname_prefix and routes_pathname_prefix default to url_base_pathname.

This launches the app at `http://0.0.0.0:8050/admin/` instead of `http://0.0.0.0:8050/`

>>Dash is running on http://0.0.0.0:8050/admin/

However, this continued to fail on deployment with the error
"The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again." aka 404

it seems like it is getting routed correctly to the dash app since https://openpath-stage.nrel.gov/admin1/ gives a different 404 and is redirected to https://www.nrel.gov/notfound

https://openpath-stage.nrel.gov/admin/ was still giving us a different 404, so presumably from within the dash app

On further investigation, this was because the app was published at http://0.0.0.0:8050/admin/, but the reverse proxy was still redirecting to proxy_pass http://[continername]:8050/;

So we ended up with

INFO:dash.dash:Dash is running on http://0.0.0.0:8050/admin/
INFO:werkzeug:192.168.1.73 - - [13/Apr/2023 02:30:46] "�[33mGET / HTTP/1.1�[0m" 404 -

Anyway, so I attempted to fix this by changing ngnix so that it would redirect to `proxy_pass http://[containername]:8050/admin;`

We are now getting an infinite redirect between http and https

![image](https://user-images.githubusercontent.com/2423263/232160967-617437fc-b0d9-44bf-9469-aebc64f1abde.png)

shankari:
Incidentally, the web app logs show a consistent 308 redirection

```
[...]

PR #25, opened 2023-04-14
Revert prior changes

shankari:
Revert changes from #23 since, per #24 they don't work


shankari:
Reverted via force push

PR #28, opened 2023-04-15
Support and test the reverse proxy configuration

shankari:
- Check in a simple docker-compose to run the app behind an ngnix proxy
- Set the relative path prefix to `/admin/`
- change all the hrefs to use relative paths


shankari:
This fixes #24

Issue #29, opened 2023-04-15
Switch to the timeseries to read data instead of accessing mongodb directly


shankari:
I had flagged this initially as a "best practice" 
https://github.com/e-mission/op-admin-dashboard/pull/7#discussion_r1146866244
https://github.com/e-mission/op-admin-dashboard/pull/7#discussion_r1142877187

However, this turned into a P0 because AWS DocumentDB (which NREL uses as a managed instance of MongoDB, and is close but not 100% compatible with it) does not seems to support the projections that are used.

We get the error

```
File "/root/miniconda-4.12.0/envs/emission/lib/python3.7/site-packages/dash/_callback.py", line 447, in add_context
output_value = func(*func_args, **func_kwargs) # %% callback invoked %%
File "app_sidebar_collapsible.py", line 190, in update_store_trips
df = query_confirmed_trips(start_date_obj, end_date_obj)
File "/usr/src/app/utils/db_utils.py", line 66, in query_confirmed_trips
df = pd.json_normalize(list(query_result))
File "/root/miniconda-4.12.0/envs/emission/lib/python3.7/site-packages/pymongo/cursor.py", line 1207, in next
if len(self.__data) or self._refresh():
File "/root/miniconda-4.12.0/envs/emission/lib/python3.7/site-packages/pymongo/cursor.py", line 1124, in _refresh
self.__send_message(q)
File "/root/miniconda-4.12.0/envs/emission/lib/python3.7/site-packages/pymongo/cursor.py", line 1001, in __send_message
address=self.__address)
File "/root/miniconda-4.12.0/envs/emission/lib/python3.7/site-packages/pymongo/mongo_client.py", line 1372, in _run_operation_with_response
exhaust=exhaust)
File "/root/miniconda-4.12.0/envs/emission/lib/python3.7/site-packages/pymongo/mongo_client.py", line 1471, in _retryable_read
return func(session, server, sock_info, slave_ok)
File "/root/miniconda-4.12.0/envs/emission/lib/python3.7/site-packages/pymongo/mongo_client.py", line 1366, in _cmd
unpack_res)
File "/root/miniconda-4.12.0/envs/emission/lib/python3.7/site-packages/pymongo/server.py", line 137, in run_operation_with_response
first, sock_info.max_wire_version)
File "/root/miniconda-4.12.0/envs/emission/lib/python3.7/site-packages/pymongo/helpers.py", line 168, in _check_command_response
max_wire_version)
[...]

PR #30, opened 2023-04-15
Hack to use the timeseries calls to retrieve the data

shankari:
This is a hack to get the dashboard to work.
It retrieves data using the timeseries interfaces and performs the projections using pandas locally.

It is *not* complete.

Notably:
- the projections are not the same as the original projections
- the projections do not take the dynamic config into account
- we still use `edb` for some calls - notably the ones to get the uuid table; we should add decorations to the timeseries to support them going forward

This is a partial fix for https://github.com/e-mission/op-admin-dashboard/issues/29

### DIFF
--- a/utils/constants.py
+++ b/utils/constants.py
@@ -2,7 +2,7 @@ valid_trip_columns = [
     "data.source",
     "data.start_ts",
     "data.start_local_dt",
-    "data.start_fmt_tm",
+    "data.start_fmt_time",
     "data.start_place",
     "data.start_loc",
     "data.end_ts",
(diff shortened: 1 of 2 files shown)


shankari:
Screenshot 2023-04-15 at 10 27 59 AM
Screenshot 2023-04-15 at 10 29 39 AM
Screenshot 2023-04-15 at 10 29 56 AM
Screenshot 2023-04-15 at 10 30 25 AM
Screenshot 2023-04-15 at 10 30 50 AM
Screenshot 2023-04-15 at 10 31 43 AM

PR #34, opened 2023-04-18
Timeseries query

AlirezaRa94:
Using time series queries to address https://github.com/e-mission/op-admin-dashboard/issues/29

### DIFF
--- a/utils/db_utils.py
+++ b/utils/db_utils.py
@@ -1,14 +1,14 @@
+import logging
 from datetime import datetime, timezone
-import sys
 
 import pandas as pd
-import logging
 
 import emission.core.get_database as edb
 import emission.storage.timeseries.abstract_timeseries as esta
 import emission.storage.timeseries.timequery as estt
 
-from utils.permissions import get_trips_columns, get_additional_trip_columns
+from utils import constants
+from utils.permissions import get_all_trip_columns, get_all_named_trip_columns
 
 
 def query_uuids(start_date, end_date):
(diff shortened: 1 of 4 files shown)


utils/db_utils.py
```diff
@@ -43,55 +43,32 @@ def query_uuids(start_date, end_date):
...
-        if 'data.end_place' in df.columns:
-            df['data.end_place'] = df['data.end_place'].apply(str)
+        columns = [col for col in get_all_trip_columns() if col in df.columns]
+        df = df[columns]
+        for col in constants.BINARY_TRIP_COLS:
+            if col in df.columns:
+                df[col] = df[col].apply(str)
+        for named_col in get_all_named_trip_columns():
+            if named_col['path'] in df.columns:
+                df[named_col['label']] = df[named_col['path']]
```
shankari, line 71: don't you have to drop the `df[named_col['path']]` columns after this? Otherwise, I think you will end up with both the label and the path as columns.

shankari (changes requested):
@AlirezaRa94 LGTM except for the comment below.
Is there a reason why this is a draft?
Can you indicate the testing done, so that I can merge it as well?

utils/db_utils.py
AlirezaRa94: That's right, addressed on https://github.com/e-mission/op-admin-dashboard/pull/34/commits/d977851530c2b6e23ef1978f3643befb93bad26a.

shankari:
I don't see any high level issues, and am happy to merge this once you have verified that it works
(aka indicated testing done)

AlirezaRa94:
> I don't see any high level issues, and am happy to merge this once you have verified that it works (aka indicated testing done)

I have tested the dashboard with the development docker-compose on both types of authentication and didn't see any errors on different pages.

shankari:
@AlirezaRa94 have you tested it with the use case of invalid columns?
e.g. what if the user specifies a column that doesn't exist in the data?
This can happen due to misconfiguration, or even for properly configured systems, before the users have entered any labels

AlirezaRa94:
> @AlirezaRa94 have you tested it with the use case of invalid columns? e.g. what if the user specifies a column that doesn't exist in the data? This can happen due to misconfiguration, or even for properly configured systems, before the users have entered any labels

Yes, It works in these scenarios, and I have tried to consider them.

shankari (approved):
Looks good in general. @AlirezaRa94 would like to see a detailed "testing done" description and testing screenshots in the future.

Issue #48, opened 2023-06-04
User screen does not show useful information about users


shankari:
To be consistent with emdash, it should show the number of trips, how many of them are confirmed, the last communication with the server... At this point, we don't see any of that information

Issue #51, opened 2023-06-05
Read only data for the past week by default


shankari:
Again, based on our experience from emdash, the default should not be to read data for a short duration (ideally one week/one month). If users choose to read more, they are welcome to.

Right now, the default is everything, and we read it as soon as the user has logged in to the dashboard. So we literally perform the most resource intensive query immediately.  When I tried this on one of the admin dashboards, both the user and the trip tables were unresponsive, and I found these errors in the console.

<img width="1792" alt="Screenshot 2023-06-04 at 5 37 57 PM" src="https://github.com/e-mission/op-admin-dashboard/assets/2423263/d296fb08-3e89-4cf9-89f7-c1995e18b9cc">

The webapp layer was essentially unresponsive.

I think that we discussed this in one of the early meetings related to the dashboard as well.

shankari:
we might also want to change all the implementations from `find_entries` to `get_data_df` as well since we only read one key at a time anyway. I don't think it will necessarily be a performance improvement but it might be with a different implementation!

shankari:
- we currently support filtering after the initial data load, but that is a problem because you cannot even see the filter range until the initial load complete
   -  this is pretty terrible from a UX perspective, because the dashboard is essentially unresponsive until the initial load is complete, which can take minutes on staging and so on
- you should never start with the computationally intensive operation and then filter down to the smaller dataset for an interactive application
   - instead, you should start with the smallest meaningful operation so you can show something immediately and then if the user chooses, allow them to start a more computationally intensive operation

which is why, I suggest reading for the past 7 days on login

Note that in this case, the initial screen that we see when we login cannot show the full sign up trend and trip trend by default because we will only load one week of trip data initially.

So we may want to think about UX design of that screen. Should we show the last week and clearly label it as "last week data" or should we just remove the trip trend and show something else meaningful instead?

I am fine with starting with something reasonable (e.g. last week trip trend) so we can get this out and then redesigning later.

Note also that not all of our tables and data are time dependent. So the participant-level data such as the user table and the demographic table need not be filtered. But the trip table and the trajectory table should be filtered by default.

I guess an alternate option might be to also read the trip table only when we access the tab. But I think that is complementary to the one week issue. Because we still may have 20,000 trips and even when we access the trip table, we don't necessarily always want to load it all and hang.

So that is a separate change, but we might as well do that when handling this scalability change.

achasmita:
[...]

PR #57, opened 2023-06-19
Add some trip stats to the user screen

AlirezaRa94:
This pull request is going to address this issue: https://github.com/e-mission/op-admin-dashboard/issues/48

### DIFF
--- a/utils/db_utils.py
+++ b/utils/db_utils.py
@@ -94,3 +94,18 @@ def query_confirmed_trips(start_date, end_date):
     # logging.debug("After filtering, the actual data is %s" % df.head())
     # logging.debug("After filtering, the actual data is %s" % df.head().trip_start_time_str)
     return df
+
+
+def add_user_stats(user_data, trip_data):
+    trip_df = pd.DataFrame(trip_data)
+    trip_grouped = trip_df.groupby('user_id')
+    for user in user_data:
+        user_id = user['user_id']
+        if user_id in trip_grouped.groups:
+            trips = trip_grouped.get_group(user_id)
+            user['confirmed_trips'] = len(trips)
+            user['first_trip'] = trips['trip_start_time_str'].min()
+            user['last_trip'] = trips['trip_start_time_str'].max()
+            # logging.debug(f"{trip_grouped.get_group(user_id).columns}")
+            # number_of_answered = len(df[df['user_input'] != {}].index) if 'user_input' in df.columns else 0
+    return user_data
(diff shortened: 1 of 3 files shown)


utils/db_utils.py
```diff
@@ -94,3 +94,18 @@ def query_confirmed_trips(start_date, end_date):
...
+    for user in user_data:
+        user_id = user['user_id']
+        if user_id in trip_grouped.groups:
+            trips = trip_grouped.get_group(user_id)
+            user['confirmed_trips'] = len(trips)
+            user['first_trip'] = trips['trip_start_time_str'].min()
+            user['last_trip'] = trips['trip_start_time_str'].max()
+            # logging.debug(f"{trip_grouped.get_group(user_id).columns}")
+            # number_of_answered = len(df[df['user_input'] != {}].index) if 'user_input' in df.columns else 0
+    return user_data
```
shankari, line 111: I am not sure that this is the desired implementation.

This may have the desired behavior at this point, but I believe it will no longer do so once https://github.com/e-mission/op-admin-dashboard/issues/51 is implemented

@asiripanich can confirm, but I believe that the goal is for the user table to include the summary of the entire trip history for the user. If we only load a week at a time, this will not have enough information.

Please use `get_first_value_for_field` to get the first and last trips instead - there should be examples of how it is used in the codebase

utils/db_utils.py
```diff
@@ -94,3 +94,18 @@ def query_confirmed_trips(start_date, end_date):
...
+
+
+def add_user_stats(user_data, trip_data):
+    trip_df = pd.DataFrame(trip_data)
+    trip_grouped = trip_df.groupby('user_id')
+    for user in user_data:
+        user_id = user['user_id']
+        if user_id in trip_grouped.groups:
+            trips = trip_grouped.get_group(user_id)
+            user['confirmed_trips'] = len(trips)
```
shankari, line 106: This seems to be `total_trips`.
`confirmed_trips` aka `labeled_trips` should be the number of trips that have labels.
I have existing script that essentially generates that this table needs
https://github.com/e-mission/e-mission-server/blob/master/bin/monitor/check_participant_status.py

Unfortunately, it uses `edb` directly, so please import `edb` in this function only (so you are not tempted to use it elsewhere) and file a new issue to add timeseries support for the functionality needed here.

shankari (changes requested):

shankari:
@AlirezaRa94 
This is good as a first step and I am fine with merging as an incremental improvement if you mark it as "ready for review".
However, to fully fix #48, we still need to add the profile information on the platform, app version, os version, etc.

shankari (approved):

AlirezaRa94:
@shankari
For now, I've used edb to get profile information, and we can change it later when needed queries are added.

PR #61, opened 2023-08-11
Feat/Segment trip time page

TTalex:
Hey,

This feature adds a new page helping users compute average trip duration between two selected points.

I wanted to experiment with a simple [level of service](https://en.wikipedia.org/wiki/Level_of_service_(transportation)) indicator based on e-mission data.

The use case comes from feedback from a local authority in France. They expressed a need for alternative ways of gathering travel time information that doesn't rely on buying data from historic actors.

I do believe that e-mission can be pretty good for this, since trip completeness isn't required to compute average durations. We can get pretty good results with low amounts of data.

The initial idea sparked a discussion around map matching. While results would probably be more accurate with map matching, [simple curve-fitting](https://github.com/MobilityNet/mobilitynet-analysis-scripts/blob/master/exploration_trajectory_timestamp_many_alternatives.ipynb), as done when creating `analysis/recreated_locations` entries, seems to already be performing well enough.

## User interaction
Here is a quick demo of the new page (the database only has one user data, mine).
segment_trip_time_demo

The user is asked to
* Select a range around start and end points
* Select a start point using the left map
* Select an end point using the right map

Using this information, queries are made to fetch recreated_locations matching either points. Resulting data is then displayed.

The proof of concept includes the following tables:
* Median segment duration by mode of transport : This is the average time taken from start to end, divided by mode of transportation
* Median segment duration by mode and hour of the day (UTC): Same as before, also divided by hour. This helps to identify peak congestion times (usually during morning and evening commute for drivers)
* Median segment duration by hour of the day (UTC): Same, without mode
* Median segment duration by mode and month: This is especially helpful to measure the impact of holidays
* Full trip data: Allows the user to build her own tables and visualizations

More useful stats could be added in the future, for example separating between weekdays and weekends.

A commonly used statistic is average vehicle speed. This requires information on true distance travelled. However, distance is complex to compute with recreated_locations, since we only fetch start and end points, we lose the distance information in intermediary points. This would require further queries to sum all intermediary points distances. The same is true for speeds. This is not complex, but could be heavy on the database / memory. In reality, the distance is most likely already known by the user, at least for the "usual" path.

## New project requirements
I have added the `dash_leaflet` library because built-in mapbox plots aren't great:
* Click events only work on markers, you cannot detect clicks on an empty map
* Drawing circles with a range in meters is not easy. Markers circles are in pixels and ignore zoom, so the solution is to rebuild a polyline with 360 points making a circle... where leaflet has a built-in Circle method

## Permissions
Three configurations are linked to this new page:
- `segment_trip_time`: User can view this page. (default `true`)
- `segment_trip_time_full_trips`: User can see the table containing non-aggregated data (default `true`)
- `segment_trip_time_min_users`: Minimal number of distinct users in data required to display anything (value is a number, default `0`). This parameter should help with guaranteeing some kind of anonymity, otherwise a user could target a specific house as a start point, and leak personal travel data that way.

## Dev notes
Up until now, the admin-dashboard was querying data once at startup. This PR behaves differently, with database queries on user actions. The code includes a few comments on why this was done this way, and the performances implications.

Hope this is useful for someone else :)

### DIFF
--- a/pages/segment_trip_time.py
+++ b/pages/segment_trip_time.py
@@ -1,0 +1,326 @@
+from dash import dcc, html, Input, Output, State, callback, register_page, dash_table
+import dash_bootstrap_components as dbc
+import dash_leaflet as dl
+import pandas as pd
+
+import emission.core.wrapper.modeprediction as ecwm
+import logging
+
+from utils.permissions import has_permission, permissions
+from utils import db_utils
+
+register_page(__name__, path="/segment_trip_time")
+
+intro = """
+## Segment average trip time
+This page displays some statistics on average trip duration between two selected points.
+"""
+
+first_step = """
+### First, select a detection radius.
+This is the range in meters around start and end point where GPS data can be detected, closest data is prioritized.
+
+This should somewhat match the distance a vehicle can cross at the maximum allowed speed in 30 seconds (sample rate).
+
+A bigger number means more trips will be considered, however it might find trips on the wrong road if roads are close enough.
+"""
+
+second_step = """
+### Then, select start and end points
+These can be anywhere on the map, but are usually on road intersections.
+
+A circle will be shown with the detection radius set on the previous step. For accurate results, the circle should not cover more than one parallel road.
+"""
+
+not_enough_data_message = f"""
+Not enough segments could be found between endpoints. This means that the number of recorded trips going from start to end point is too low. 
+* There could be data, but on an insufficient number of users, breaking anonymity (minimum number of users is currently set to {permissions.get('segment_trip_time_min_users', 0)})
+* You could try to increase the detection radius, or chose different start and end points.
+"""
+
+layout = html.Div(
+    [
+        dcc.Store(id='link-trip-time-start', data=(0, 0)),
+        dcc.Store(id='link-trip-time-end', data=(0, 0)),
+        # dcc.Store(id='store-mode-by-section-id', data={}),
+        dcc.Markdown(intro),
+        dcc.Markdown(first_step),
+        dcc.Slider(
+            0,
+            2500,
+            id='detection-radius',
+            value=200,
+            tooltip={"placement": "bottom", "always_visible": True},
+        ),
+        dcc.Markdown(second_step),
+        dbc.Row(
+            [
+                dbc.Col(
+                    [
+                        html.H4('Start point selection'),
+                        dl.Map(
+                            [dl.TileLayer(), dl.LayerGroup(id='stt-trip-layer-start')],
+                            id='stt-trip-map-start',
+                            style={'height': '500px'},
+                            center=[32.7, -96.8],
+                            zoom=5,
+                        ),
+                    ]
+                ),
+                dbc.Col(
+                    [
+                        html.H4('End point selection'),
+                        dl.Map(
+                            [dl.TileLayer(), dl.LayerGroup(id='stt-trip-layer-end')],
+                            id='stt-trip-map-end',
+                            style={'height': '500px'},
+                            center=[32.7, -96.8],
+                            zoom=5,
+                        ),
+                    ]
+                ),
+            ]
+        ),
+        dbc.Row(
+            html.Div(id='message'),
+        ),
+    ]
+)
+
+
+def map_click(click_lat_lng, radius, circle_color):
+    # WARN: There seem to be a bug on click_lat_lng leaflet event, values can be out of bound (e.g: -357 for lat in eu), some kind of modulus might be required
+    # Couldn't reproduce it though
+    layer_children = [
+        dl.Circle(center=click_lat_lng, radius=radius, color=circle_color)
+    ]
+    endpoint_coords = click_lat_lng
+    zoom = 13
+    map_center = click_lat_lng
+    return layer_children, endpoint_coords, zoom, map_center
+
+
+@callback(
+    Output('stt-trip-map-end', 'zoom', allow_duplicate=True),
+    Output('stt-trip-map-end', 'center', allow_duplicate=True),
+    Input('link-trip-time-start', 'data'),
+    State('stt-trip-map-end', 'zoom'),
+    State('stt-trip-map-end', 'center'),
+    State('link-trip-time-end', 'data'),
+    prevent_initial_call=True,
+)
+# This is optional, it's used on first selection to help user locate his endpoint
+def center_end_map_helper(link_trip_time_start, end_zoom, end_center, link_trip_time_end):
+    if link_trip_time_end[0] == 0:
+        end_zoom = 13
+        end_center = link_trip_time_start
+    return end_zoom, end_center
+
+
+@callback(
+    Output('stt-trip-layer-start', 'children'),
+    Output('link-trip-time-start', 'data'),
+    Output('stt-trip-map-start', 'zoom'),
+    Output('stt-trip-map-start', 'center'),
+    Input('stt-trip-map-start', 'click_lat_lng'),
+    State('detection-radius', 'value'),
+    prevent_initial_call=True,
+)
+def map_start_click(click_lat_lng, radius):
+    return map_click(click_lat_lng, radius, "green")
+
+
+@callback(
+    Output('stt-trip-layer-end', 'children'),
+    Output('link-trip-time-end', 'data'),
+    Output('stt-trip-map-end', 'zoom'),
+    Output('stt-trip-map-end', 'center'),
+    Input('stt-trip-map-end', 'click_lat_lng'),
+    State('detection-radius', 'value'),
+    prevent_initial_call=True,
+)
+def map_end_click(click_lat_lng, radius):
+    return map_click(click_lat_lng, radius, "red")
+
+
+def format_duration_df(df, time_column_name='Time sample'):
+    df['Median time (minutes)'] = df.duration / 60  # convert seconds in minutes
+    df = df.reset_index().rename(
+        columns={
+            'start_fmt_time': time_column_name,
+            'duration': 'Median time (seconds)',
+            'section': 'Count',
+            'mode': 'Mode',
+        }
+    )
+    if time_column_name in df:
+        if 'Mode' in df:
+            df = df[
+                [
+                    'Mode',
+                    time_column_name,
+                    'Median time (seconds)',
+                    'Median time (minutes)',
+                    'Count',
+                ]
+            ]  # reorder cols
+        else:
+            df = df[
+                [
+                    time_column_name,
+                    'Median time (seconds)',
+                    'Median time (minutes)',
+                    'Count',
+                ]
+            ]  # reorder cols
+    else:
+        df = df[
+            ['Mode', 'Median time (seconds)', 'Median time (minutes)', 'Count']
+        ]  # reorder cols
+    df = df.to_dict('records')  # Format for display
+    return df
+
+
+@callback(
+    Output('message', 'children'),
+    Input('link-trip-time-start', 'data'),
+    Input('link-trip-time-end', 'data'),
+    State('detection-radius', 'value'),
+    prevent_initial_call=True,
+)
+def generate_content_on_endpoints_change(link_trip_time_start, link_trip_time_end, radius):
+    if link_trip_time_end[0] == 0 or link_trip_time_start[0] == 0:
+        return ''
+    # logging.debug("link_trip_time_start: " + str(link_trip_time_start))
+    # logging.debug("link_trip_time_end: " + str(link_trip_time_end))
+
+    # Warning: This is a database call, looks here if there is a performance hog.
+    # From initial tests, this seems to be performing well, without the need to do geoqueries in memory
+    df = db_utils.query_segments_crossing_endpoints(
+        link_trip_time_start[0],
+        link_trip_time_start[1],
+        link_trip_time_end[0],
+        link_trip_time_end[1],
+        radius,
+    )
+    total_nb_trips = df.shape[0]
+    if total_nb_trips > 0:
+        modes = [e.name for e in ecwm.PredictedModeTypes]
+        # Warning: Another db call here.
+        # In theory, we could load all inferred_section modes in memory at start time, instead of fetching it everytime
+        # However, when testing it, the operation is quite heavy on the db and on ram.
+        # I opted for querying only sections we're interested in, every time. Page load is still decent, especially when the number of section is low.
+        mode_by_section_id = db_utils.query_inferred_sections_modes(
+            df['section'].to_list()
+        )
+        df['mode'] = df['section'].apply(
+            lambda section_id: modes[mode_by_section_id[str(section_id)]]
+        )
+        median_trip_time = df['duration'].median()
+        times = pd.to_datetime(df['start_fmt_time'], errors='coerce', utc=True)
+        duration_per_hour = format_duration_df(
+            df.groupby(times.dt.hour).agg({'duration': 'median', 'section': 'count'}),
+            time_column_name='Hour',
+        )
+        duration_per_mode = format_duration_df(
+            df.groupby('mode').agg({'duration': 'median', 'section': 'count'})
+        )
+        duration_per_mode_per_hour = format_duration_df(
+            df.groupby(['mode', times.dt.hour]).agg(
+                {'duration': 'median', 'section': 'count'}
+            ),
+            time_column_name='Hour',
+        )
+        duration_per_mode_per_month = format_duration_df(
+            df.groupby(['mode', times.dt.month]).agg(
+                {'duration': 'median', 'section': 'count'}
+            ),
+            time_column_name='Month',
+        )
+        return dbc.Row(
+            [
+                dbc.Col(
+                    [
+                        html.Br(),
+                        html.H3('Results'),
+                        html.Div(
+                            f'Computed median segment duration is {median_trip_time} seconds, {total_nb_trips} trips considered'
+                        ),
+                        html.Br(),
+                        html.H4('Median segment duration by mode of transport'),
+                        dash_table.DataTable(
+                            id='duration_per_mode',
+                            data=duration_per_mode,
+                            sort_action='native',
+                            sort_mode='multi',
+                            export_format='csv',
+                        ),
+                        html.Br(),
+                        html.H4(
+                            'Median segment duration by mode and hour of the day (UTC)'
+                        ),
+                        dash_table.DataTable(
+                            id='duration_per_hour',
+                            data=duration_per_hour,
+                            sort_action='native',
+                            sort_mode='multi',
+                            export_format='csv',
+                        ),
+                        html.Br(),
+                        html.H4(
+                            'Median segment duration by mode and hour of the day (UTC)'
+                        ),
+                        dash_table.DataTable(
+                            id='duration_per_mode_per_hour',
+                            data=duration_per_mode_per_hour,
+                            sort_action='native',
+                            sort_mode='multi',
+                            export_format='csv',
+                        ),
+                        html.Br(),
+                        html.H4('Median segment duration by mode and month'),
+                        dash_table.DataTable(
+                            id='duration_per_mode_per_month',
+                            data=duration_per_mode_per_month,
+                            sort_action='native',
+                            sort_mode='multi',
+                            export_format='csv',
+                        ),
+                    ],
+                    xs=6,
+                ),
+                dbc.Col(
+                    [
+                        html.Br(),
+                        html.H3('Trips Data'),
+                        dash_table.DataTable(
+                            id='trips_data',
+                            data=df[
+                                ['start_fmt_time', 'end_fmt_time', 'mode', 'duration']
+                            ].to_dict('records'),
+                            page_size=15,
+                            sort_action='native',
+                            sort_mode='multi',
+                            export_format='csv',
+                        ),
+                    ],
+                    xs=6,
+                    style={
+                        'display': 'block'
+                        if has_permission('segment_trip_time_full_trips')
+                        else 'none'
+                    },
+                ),
+            ]
+        )
+    return [html.H3('Results'), dcc.Markdown(not_enough_data_message)]
+
+
+# This is left as an example on loading all inferred_section modes in memory at start time, instead of fetching it everytime
+# This lead to poor memory performances on a larger db
+# @callback(
+#     Output('store-mode-by-section-id', 'data'),
+#     Input('store-trips', 'data') # Only using this an initial trigger
+# )
+# def load_mode_by_section_id(s):
+#     return db_utils.query_inferred_sections_modes()
--- a/pages/segment_trip_time.py
+++ b/pages/segment_trip_time.py
@@ -3,7 +3,6 @@ import dash_bootstrap_components as dbc
 import dash_leaflet as dl
 import pandas as pd
 
-import emission.core.wrapper.modeprediction as ecwm
 import logging
 import json
 
@@ -175,16 +174,15 @@ def generate_content_on_endpoints_change(link_trip_time_start_str, link_trip_tim
     )
     total_nb_trips = df.shape[0]
     if total_nb_trips > 0:
-        modes = [e.name for e in ecwm.PredictedModeTypes]
         # Warning: Another db call here.
         # In theory, we could load all inferred_section modes in memory at start time, instead of fetching it everytime
         # However, when testing it, the operation is quite heavy on the db and on ram.
         # I opted for querying only sections we're interested in, every time. Page load is still decent, especially when the number of section is low.
         mode_by_section_id = db_utils.query_inferred_sections_modes(
-            df['section'].to_list()
+            df[['section', 'user_id']].to_dict('records')
         )
         df['mode'] = df['section'].apply(
-            lambda section_id: modes[mode_by_section_id[str(section_id)]]
+            lambda section_id: mode_by_section_id[str(section_id)].name
         )
         median_trip_time = df['duration'].median()
         times = pd.to_datetime(df['start_fmt_time'], errors='coerce', utc=True)
@@ -284,13 +282,3 @@ def generate_content_on_endpoints_change(link_trip_time_start_str, link_trip_tim
             ]
         )
     return [html.H3('Results'), dcc.Markdown(not_enough_data_message)]
-
-
-# This is left as an example on loading all inferred_section modes in memory at start time, instead of fetching it everytime
-# This lead to poor memory performances on a larger db
-# @callback(
-#     Output('store-mode-by-section-id', 'data'),
-#     Input('store-trips', 'data') # Only using this an initial trigger
-# )
-# def load_mode_by_section_id(s):
-#     return db_utils.query_inferred_sections_modes()
--- a/pages/segment_trip_time.py
+++ b/pages/segment_trip_time.py
@@ -257,7 +257,7 @@ def generate_content_on_endpoints_change(link_trip_time_start, link_trip_time_en
                         ),
                         html.Br(),
                         html.H4(
-                            'Median segment duration by mode and hour of the day (UTC)'
+                            'Median segment duration by hour of the day (UTC)'
                         ),
                         dash_table.DataTable(
                             id='duration_per_hour',
--- a/pages/segment_trip_time.py
+++ b/pages/segment_trip_time.py
@@ -5,6 +5,7 @@ import pandas as pd
 
 import emission.core.wrapper.modeprediction as ecwm
 import logging
+import json
 
 from utils.permissions import has_permission, permissions
 from utils import db_utils
@@ -13,69 +14,76 @@ register_page(__name__, path="/segment_trip_time")
 
 intro = """
 ## Segment average trip time
-This page displays some statistics on average trip duration between two selected points.
+This page displays some statistics on average trip duration between two selected zones.
+
+### Usage
+Using the polygon or square tools on the maps' menu, draw the start (left map) and end (right map) zones to consider.
+
+Data will then be fetched for trips crossing the start zone and then the end zone.
+
+Here are some tips on how to draw zones:
+* Zones shouldn't cover more than one parallel road; otherwise, it is unclear which path the user took.
+* A bigger zone will give more results, at the cost of lower accuracy in trip durations (the start point could be anywhere in the zone).
+* For exhaustivity, zone length should somewhat match the distance a vehicle can cross at the maximum allowed speed in 30 seconds (sample rate).
+* A smaller zone will give more accurate time results, but the number of trips might be too low to be significant.
+* Zones can be moved and edited using the Edit layer menu, and they can be deleted with the Delete layer button.
+* Please be advised that only the last added zone will be considered on each map. It is thus advised to delete existing zones before creating new ones.
 """
 
-first_step = """
-### First, select a detection radius.
-This is the range in meters around start and end point where GPS data can be detected, closest data is prioritized.
-
-This should somewhat match the distance a vehicle can cross at the maximum allowed speed in 30 seconds (sample rate).
-
-A bigger number means more trips will be considered, however it might find trips on the wrong road if roads are close enough.
-"""
-
-second_step = """
-### Then, select start and end points
-These can be anywhere on the map, but are usually on road intersections.
-
-A circle will be shown with the detection radius set on the previous step. For accurate results, the circle should not cover more than one parallel road.
-"""
 
 not_enough_data_message = f"""
 Not enough segments could be found between endpoints. This means that the number of recorded trips going from start to end point is too low. 
 * There could be data, but on an insufficient number of users, breaking anonymity (minimum number of users is currently set to {permissions.get('segment_trip_time_min_users', 0)})
-* You could try to increase the detection radius, or chose different start and end points.
+* You could try to increase the zone sizes, or chose different start and end points.
 """
 
+initial_maps_center = [32.7, -96.8]
+initial_maps_zoom = 5
 layout = html.Div(
     [
-        dcc.Store(id='link-trip-time-start', data=(0, 0)),
-        dcc.Store(id='link-trip-time-end', data=(0, 0)),
-        # dcc.Store(id='store-mode-by-section-id', data={}),
+        dcc.Store(id='link-trip-time-start', data=json.dumps({"features": []})),
+        dcc.Store(id='link-trip-time-end', data=json.dumps({"features": []})),
         dcc.Markdown(intro),
-        dcc.Markdown(first_step),
-        dcc.Slider(
-            0,
-            2500,
-            id='detection-radius',
-            value=200,
-            tooltip={"placement": "bottom", "always_visible": True},
-        ),
-        dcc.Markdown(second_step),
         dbc.Row(
             [
                 dbc.Col(
                     [
-                        html.H4('Start point selection'),
+                        html.H4('Start zone selection'),
                         dl.Map(
-                            [dl.TileLayer(), dl.LayerGroup(id='stt-trip-layer-start')],
+                            [
+                                dl.TileLayer(), 
+                                dl.FeatureGroup([
+                                    dl.EditControl(
+                                        id="stt-edit-control-start", 
+                                        draw=dict(circle=False, marker=False, polyline=False, circlemarker=False)
+                                    )
+                                ])
+                            ],
+                            #[dl.TileLayer(), dl.LayerGroup(id='stt-trip-layer-start')],
                             id='stt-trip-map-start',
                             style={'height': '500px'},
-                            center=[32.7, -96.8],
-                            zoom=5,
+                            center=initial_maps_center,
+                            zoom=initial_maps_zoom
                         ),
                     ]
                 ),
                 dbc.Col(
                     [
-                        html.H4('End point selection'),
+                        html.H4('End zone selection'),
                         dl.Map(
-                            [dl.TileLayer(), dl.LayerGroup(id='stt-trip-layer-end')],
+                            [
+                                dl.TileLayer(), 
+                                dl.FeatureGroup([
+                                    dl.EditControl(
+                                        id="stt-edit-control-end", 
+                                        draw=dict(circle=False, marker=False, polyline=False, circlemarker=False)
+                                    )
+                                ])
+                            ],
                             id='stt-trip-map-end',
                             style={'height': '500px'},
-                            center=[32.7, -96.8],
-                            zoom=5,
+                            center=initial_maps_center,
+                            zoom=initial_maps_zoom
                         ),
                     ]
                 ),
@@ -88,59 +96,23 @@ layout = html.Div(
 )
 
 
-def map_click(click_lat_lng, radius, circle_color):
-    # WARN: There seem to be a bug on click_lat_lng leaflet event, values can be out of bound (e.g: -357 for lat in eu), some kind of modulus might be required
-    # Couldn't reproduce it though
-    layer_children = [
-        dl.Circle(center=click_lat_lng, radius=radius, color=circle_color)
-    ]
-    endpoint_coords = click_lat_lng
-    zoom = 13
-    map_center = click_lat_lng
-    return layer_children, endpoint_coords, zoom, map_center
-
 
 @callback(
-    Output('stt-trip-map-end', 'zoom', allow_duplicate=True),
-    Output('stt-trip-map-end', 'center', allow_duplicate=True),
-    Input('link-trip-time-start', 'data'),
-    State('stt-trip-map-end', 'zoom'),
-    State('stt-trip-map-end', 'center'),
-    State('link-trip-time-end', 'data'),
-    prevent_initial_call=True,
-)
-# This is optional, it's used on first selection to help user locate his endpoint
-def center_end_map_helper(link_trip_time_start, end_zoom, end_center, link_trip_time_end):
-    if link_trip_time_end[0] == 0:
-        end_zoom = 13
-        end_center = link_trip_time_start
-    return end_zoom, end_center
-
-
-@callback(
-    Output('stt-trip-layer-start', 'children'),
     Output('link-trip-time-start', 'data'),
-    Output('stt-trip-map-start', 'zoom'),
-    Output('stt-trip-map-start', 'center'),
-    Input('stt-trip-map-start', 'click_lat_lng'),
-    State('detection-radius', 'value'),
+    Input('stt-edit-control-start', 'geojson'),
     prevent_initial_call=True,
 )
-def map_start_click(click_lat_lng, radius):
-    return map_click(click_lat_lng, radius, "green")
-
+def map_start_draw(geojson):
+    return json.dumps(geojson)
 
 @callback(
-    Output('stt-trip-layer-end', 'children'),
     Output('link-trip-time-end', 'data'),
-    Output('stt-trip-map-end', 'zoom'),
-    Output('stt-trip-map-end', 'center'),
-    Input('stt-trip-map-end', 'click_lat_lng'),
-    State('detection-radius', 'value'),
+    Input('stt-edit-control-end', 'geojson'),
     prevent_initial_call=True,
 )
-def map_end_click(click_lat_lng, radius):
-    return map_click(click_lat_lng, radius, "red")
+def map_end_draw(geojson):
+    return json.dumps(geojson)
+
 
 
 def format_duration_df(df, time_column_name='Time sample'):
@@ -185,11 +157,12 @@ def format_duration_df(df, time_column_name='Time sample'):
     Output('message', 'children'),
     Input('link-trip-time-start', 'data'),
     Input('link-trip-time-end', 'data'),
-    State('detection-radius', 'value'),
     prevent_initial_call=True,
 )
-def generate_content_on_endpoints_change(link_trip_time_start, link_trip_time_end, radius):
-    if link_trip_time_end[0] == 0 or link_trip_time_start[0] == 0:
+def generate_content_on_endpoints_change(link_trip_time_start_str, link_trip_time_end_str):
+    link_trip_time_start = json.loads(link_trip_time_start_str)
+    link_trip_time_end = json.loads(link_trip_time_end_str)
+    if len(link_trip_time_end["features"]) == 0 or len(link_trip_time_start["features"]) == 0:
         return ''
     # logging.debug("link_trip_time_start: " + str(link_trip_time_start))
     # logging.debug("link_trip_time_end: " + str(link_trip_time_end))
@@ -197,11 +170,8 @@ def generate_content_on_endpoints_change(link_trip_time_start, link_trip_time_en
     # Warning: This is a database call, looks here if there is a performance hog.
     # From initial tests, this seems to be performing well, without the need to do geoqueries in memory
     df = db_utils.query_segments_crossing_endpoints(
-        link_trip_time_start[0],
-        link_trip_time_start[1],
-        link_trip_time_end[0],
-        link_trip_time_end[1],
-        radius,
+        link_trip_time_start["features"][len(link_trip_time_start["features"])-1],
+        link_trip_time_end["features"][len(link_trip_time_end["features"])-1],
     )
     total_nb_trips = df.shape[0]
     if total_nb_trips > 0:
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,4 +14,4 @@ python-jose==3.3.0
 flask==2.2.5
 flask-talisman==1.0.0
 dash_auth==2.0.0
-dash-leaflet==0.1.28
+dash-leaflet==1.0.7
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,11 +15,3 @@ flask==2.2.5
 flask-talisman==1.0.0
 dash_auth==2.0.0
 dash-leaflet==0.1.28
-
-# emission requirements
-arrow==1.2.3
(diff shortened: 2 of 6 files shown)


shankari:
@TTalex I think that this is a great feature, and super well documented! We just deployed a release, but are planning a new one in a couple of weeks for the API upgrade, and it would be great to include this as well. I'll review and merge this weekend.

shankari:
FYI, map matching has been postponed in favor of helping finish up "count every trip"

pages/segment_trip_time.py
```diff
@@ -0,0 +1,326 @@
+from dash import dcc, html, Input, Output, State, callback, register_page, dash_table
+import dash_bootstrap_components as dbc
+import dash_leaflet as dl
```
shankari, line 3: I agree that the mapbox maps that we have are currently unintuitive. They try to fit the plot interface with pan and zoom, but it always ends up centering me when I am trying to zoom around and then I have to go zoom out and find where I was again.

shankari: @achasmita, we should probably replace the other mapbox maps with leaflet as well.

utils/db_utils.py
```diff
@@ -158,4 +159,87 @@ def add_user_stats(user_data):
...
+                'end_fmt_time': matching_end_data.get('fmt_time')
+            })
+    if perm_utils.permissions.get("segment_trip_time_min_users", 0) <= number_user_seen:
+        return pd.DataFrame.from_dict(vals)
+    return pd.DataFrame.from_dict([])
+
+# The following query can be called multiple times, so we extract index creation from the function
+analysis_timeseries_db = edb.get_analysis_timeseries_db()
+# In theory, adding an index on the cleaned_section should improve query performance
+analysis_timeseries_db.create_index([('data.cleaned_section', pymongo.ASCENDING)])
```
shankari, line 229: index creation in mongodb is not temporary.
We should put this into `emission.core.database` just like all the other index calls.

utils/db_utils.py
```diff
@@ -158,4 +159,87 @@ def add_user_stats(user_data):
...
                 user['last_call'] = arrow.get(last_call).format(time_format)
 
-    return user_data
\ No newline at end of file
+    return user_data
+
+def query_segments_crossing_endpoints(start_lat, start_long, end_lat, end_long, range_around_endpoints=400):
+    # data.loc only appears in analysis/recreated_location
+    query_start = {'data.loc': {'$near': SON([('$geometry', SON([('type', 'Point'), ('coordinates', [start_long, start_lat])])), ('$maxDistance', range_around_endpoints)])}}
+    query_end = {'data.loc': {'$near': SON([('$geometry', SON([('type', 'Point'), ('coordinates', [end_long, end_lat])])), ('$maxDistance', range_around_endpoints)])}}
```
shankari, line 167: it seems like it would be better to use `emission/storage/timeseries/geoquery.py` with the timeseries in `emission/storage/timeseries`. That will make it easier to migrate to a different DB/storage solution if we want to later.

Note that this container is built on the e-mission-server base container and has access to the full e-mission-server codebase.

I do note that the current implementation of `emission/storage/timeseries/geoquery` only supports `$geoWithin` a `Polygon` or `MultiPolygon`, but it should be easy to extend. Let us know if you would like us to tackle that instead.

I'm also curious as to why you needed `SON` to get this to work - the current polygon based geoquery is used in a unit test with a simple dict, no `SON` and it seems to work.

utils/db_utils.py
```diff
@@ -158,4 +159,87 @@ def add_user_stats(user_data):
...
+    end_by_section = {}
+    for elt in res_end:
+        elt_data = elt.get('data')
+        if elt_data:
+            section_id = elt_data.get('section')
+            # This makes sure that only the first encounter of the section is added
+            # The near query gives closest points first, so the first one is the one we want
+            if section_id not in end_by_section:
+                end_by_section[section_id] = elt
+
```
shankari, line 179: you can then use `get_data_df` with the geoquery from the timeseries, and this code can collapse to something like

```
agg_ts = eastag.AggregateTimeSeries().get_aggregate_time_series()
locs_matching_end = agg_ts.get_data_df("analysis/recreated_location", geo_query = gq)
end_by_section = locs_matching_end.section_id.unique()
```
and get the benefit of the C speedup

requirements.txt
```diff
@@ -14,3 +14,12 @@ python-jose==3.3.0
...
 dash_auth==2.0.0
+dash-leaflet==0.1.28
+
+# emission requirements
+arrow==1.2.3
+pymongo==4.4.1
+pandas==2.0.3
+future==0.18.3
+attrdict==2.0.1
+geojson==3.0.1
```
shankari, line 25: Again, if you run this in the container defined in the `docker-compose.yml`, you will not need all these dependencies.

shankari (changes requested):
@TTalex this is a great PR. I just have some suggestions on how to integrate it into e-mission-server. I just realized that the README is obsolete, and the intern who submitted a doc PR left without addressing my comments

Went through and removed all the obsolete instructions:
https://github.com/e-mission/op-admin-dashboard/commit/43f740bbf8c87672bb75dd6af84a295092f0d60c

requirements.txt
TTalex: Sweet, I'll nuke these lines :+1:

utils/db_utils.py
TTalex: True, removed in commit https://github.com/e-mission/op-admin-dashboard/pull/61/commits/88c9665ff0a6f0863372e239ed8f038eadcb5770 
And added in PR https://github.com/e-mission/e-mission-server/pull/934

utils/db_utils.py
TTalex: The `SON` lib was indeed unneeded, I removed it in commit https://github.com/e-mission/op-admin-dashboard/pull/61/commits/25e51792d786541ef547f821b8cc38f15567213d

For the geoquery extension, I believe I might need a hand. I'm running into issues with `count_documents` not supporting `$near` queries. (See https://www.mongodb.com/community/forums/t/why-the-countdocuments-doesnt-support-geonear/1330)
This limitation leads to an error here: https://github.com/e-mission/e-mission-server/blob/a8c7dadac63480745c88b8b1cc643df6fc70fa7b/emission/storage/timeseries/builtin_timeseries.py#L232 that I'm not sure how to handle correctly.
`pymongo.errors.OperationFailure: $geoNear, $near, and $nearSphere are not allowed in this context, full error: {'ok': 0.0, 'errmsg': '$geoNear, $near, and $nearSphere are not allowed in this context', 'code': 2, 'codeName': 'BadValue'}`

utils/db_utils.py
TTalex: Good point, I have rewritten the method using this. It isn't yet committed, since changes will be needed depending on the GeoQuery implementation.
I'm pasting the code here for future reference
```python
def query_segments_crossing_endpoints(start_lat, start_long, end_lat, end_long, range_around_endpoints=400):
    start_gq = GeoQuery(xxx, start_long, start_lat, range_around_endpoints) # TBD
    end_gq = GeoQuery(xxx, end_long, end_lat, range_around_endpoints) # TBD
    agg_ts = estag.AggregateTimeSeries().get_aggregate_time_series()

    locs_matching_start = agg_ts.get_data_df("analysis/recreated_location", geo_query = start_gq)
    locs_matching_start = locs_matching_start.drop_duplicates(subset=['section'])
    if locs_matching_start.empty:
        return locs_matching_start # No need to continue, no results
    
    locs_matching_end = agg_ts.get_data_df("analysis/recreated_location", geo_query = end_gq)
    locs_matching_end = locs_matching_end.drop_duplicates(subset=['section'])
    if locs_matching_end.empty:
        return locs_matching_end # No need to continue, no results
    
    merged = locs_matching_start.merge(locs_matching_end, how='outer', on=['section'])
    filtered = merged.loc[merged['idx_x']<merged['idx_y']].copy()
    filtered['duration'] = filtered['ts_y'] - filtered['ts_x']
    filtered['mode'] = filtered['mode_x']
    filtered['start_fmt_time'] = filtered['fmt_time_x']
    filtered['end_fmt_time'] = filtered['fmt_time_y']
    
    number_user_seen = filtered.user_id_x.nunique()

    if perm_utils.permissions.get("segment_trip_time_min_users", 0) <= number_user_seen:
        return filtered
    return pd.DataFrame.from_dict([])
```
Given the required database calls, I do not believe C speedup will make a big difference in the whole process, but eh, might as well try to optimize.

utils/db_utils.py
shankari: @TTalex `$geoWithin` seems to work with `count_documents` (https://github.com/e-mission/e-mission-server/pull/936/commits/b6d90cfb232960a04cf93e475c5780ba9abc67a0). You should be able to use the `$center` operator as a drop-in replacement for your code. Although if you have time to investigate `leaflet.draw` (https://leaflet.github.io/Leaflet.draw/docs/leaflet-draw-latest.html) and its use in plotly, you could allow the user to draw the polygon at the start and end and then use `geoWithin` with a polygon.

Note that this likely strictly better - users can have a different radius for start and end, and they can draw their polygon to exclude intersections they don't want... But I don't know how easy that leaflet draw integration is.

Finally, please note that `$near` is less performant because it sorts the results and `$geoWithin` is recommended for better performance unless you need the sorting.
https://www.mongodb.com/docs/manual/reference/operator/query/near/

> If the order of the documents is not important to you, consider using the [$geoWithin](https://www.mongodb.com/docs/manual/reference/operator/query/geoWithin/#mongodb-query-op.-geoWithin) operator instead, as it returns unsorted results.

utils/db_utils.py
```diff
@@ -158,4 +158,104 @@ def add_user_stats(user_data):
...
+def query_inferred_sections_modes(sections=[]):
+    query = {'metadata.key': 'analysis/inferred_section'}
+    if len(sections) > 0:
+        query['data.cleaned_section'] = {'$in': sections} 
+    res = analysis_timeseries_db.find(query, {'data.cleaned_section': 1, 'data.sensed_mode': 1})
+    mode_by_section_id = {}
+    for elt in res:
+        elt_data = elt.get('data')
+        if elt_data:
+            mode_by_section_id[str(elt_data.get('cleaned_section'))] = elt_data.get('sensed_mode') or 0
```
shankari, line 260: I see your note here about performance issues while reading the inferred sections.
And I also see that you are making a direct mongoDB query from here, which we want to avoid  to make it easier to refactor the core as needed later.

I would suggest iterating through the sections and using `cleaned2inferred_section` in `emission/storage/decorations/section_queries` on each one.

As an aside, I should really think about how to fix the datamodel so that we don't need this two step retrieval and the unnecessary performance hit.

shankari (changes requested):

utils/db_utils.py
TTalex: Good insight, thanks for that.

Using draw was the right option, it's way more precise for users than circles, and it looks like it's working well with the existing Geoquery method.

I re-wrote most of the code to use this solution in commit https://github.com/e-mission/op-admin-dashboard/pull/61/commits/59e0c9810e68981593afef224a345f5c38732227

Moreover, dash-leaflet went into v1.x.x a few days ago, breaking some interactions with circles and offering a clean interaction with leflet.draw. So timing is great !

TTalex:
Quoting myself from the first message of this PR:
> A commonly used statistic is average vehicle speed. This requires information on true distance travelled. However, distance is complex to compute with recreated_locations, since we only fetch start and end points, we lose the distance information in intermediary points. This would require further queries to sum all intermediary points distances. The same is true for speeds. This is not complex, but could be heavy on the database / memory. In reality, the distance is most likely already known by the user, at least for the "usual" path.

This could be solved by having a `total_distance_from_start` field in the recreated location. **I wonder if this would be an interesting feature for other use cases ?**

It could be implemented in `add_dist_heading_speed` https://github.com/e-mission/e-mission-server/blob/f78a22b7735e3877f31b29a4f7029dbd182416d4/emission/analysis/intake/cleaning/location_smoothing.py#L71 as follows:
```diff
+ import itertools
# [...]
def add_dist_heading_speed(points_df):
    # type: (pandas.DataFrame) -> pandas.DataFrame
    """
    Returns a new dataframe with an added "speed" column.
    The speed column has the speed between each point and its previous point.
    The first row has a speed of zero.
    """
    point_list = [ad.AttrDict(row) for row in points_df.to_dict('records')]
    zipped_points_list = list(zip(point_list, point_list[1:]))

    distances = [pf.calDistance(p1, p2) for (p1, p2) in zipped_points_list]
    distances.insert(0, 0)
+   distances_from_start = list(itertools.accumulate(distances))
    speeds = [pf.calSpeed(p1, p2) for (p1, p2) in zipped_points_list]
    speeds.insert(0, 0)
    headings = [pf.calHeading(p1, p2) for (p1, p2) in zipped_points_list]
    headings.insert(0, 0)

    with_distances_df = pd.concat([points_df, pd.Series(distances, name="distance")], axis=1)
+   with_distances_from_start_df = pd.concat([with_distances_df, pd.Series(distances_from_start, name="distance_from_start")], axis=1)
-   with_speeds_df = pd.concat([with_distances_df, pd.Series(speeds, name="speed")], axis=1)
+   with_speeds_df = pd.concat([with_distances_from_start_df, pd.Series(speeds, name="speed")], axis=1)
    if "heading" in with_speeds_df.columns:
        with_speeds_df.drop("heading", axis=1, inplace=True)
    with_headings_df = pd.concat([with_speeds_df, pd.Series(headings, name="heading")], axis=1)
    return with_headings_df
```

Recreated locations would then look like the following examples:
`{metadata: {key: "analysis/recreated_location", [...]}, data: {idx: 0, distance: 0, distance_from_start:0, [...]}}`
`{metadata: {key: "analysis/recreated_location", [...]}, data: {idx: 1, distance: 105, distance_from_start:105, [...]}}`
`{metadata: {key: "analysis/recreated_location", [...]}, data: {idx: 2, distance: 65, distance_from_start:170, [...]}}`
`{metadata: {key: "analysis/recreated_location", [...]}, data: {idx: 3, distance: 100, distance_from_start:270, [...]}}`

Computing distance from the second point (idx 1) to the last one (idx 3) would only require both points, skipping fetching idx 2: `270-105=165`

Maybe it's a bit too specific for this use case to induce a change to the Location model. (and might require a patch on existing database entries for consistency 😕)

utils/db_utils.py
shankari: I have filed https://github.com/e-mission/e-mission-docs/issues/970 to create a new interface for this. Once we have the interface, we can experiment or switch implementations under the hood. My personal preference is for the timeseries model (so we find potentially matching inferred sections by querying by time range or start/end geo) but I don't know if that will be feasible for this use case - e.g. whether the query will be tight enough.

shankari:
@TTalex we have created an interface for the `cleaned2inferredsections` mapping.
Can you switch to it? LMK if you are too busy, and one of us can handle it.
https://github.com/e-mission/e-mission-server/pull/937

We can then change the implementation at will depending on the scalability vs. reuse tradeoff.

shankari:
> This could be solved by having a total_distance_from_start field in the recreated location. I wonder if this would be an interesting feature for other use cases ?

This is an interesting thought. Adding new entries to the data model and patching existing entries is work but fairly straightforward conceptually. If we can come up with a second use case that needs this functionality (maybe map matching), I am happy to include it. Not sure if we want to do a one-off change before that though...

TTalex:
> @TTalex we have created an interface for the `cleaned2inferredsections` mapping. Can you switch to it? LMK if you are too busy, and one of us can handle it. [e-mission/e-mission-server#937](https://github.com/e-mission/e-mission-server/pull/937)

Sweet, thanks, I've made the swap.

TTalex:
> This is an interesting thought. Adding new entries to the data model and patching existing entries is work but fairly straightforward conceptually. If we can come up with a second use case that needs this functionality (maybe map matching), I am happy to include it. Not sure if we want to do a one-off change before that though...

I wouldn't change it either if I were you, that's why I didn't bother doing a PR :)

There might be some use cases in end users UI for it, where the change would induce slight performance improvements. For example, with point by point visualization such as this one:

Peek 2023-10-02 11-20

But I'm not confident it would be an improvement at all, since the full list of points has to be loaded anyway.

shankari (approved):

shankari:
I tried testing it on some real-world data in Denver and only got 3 trips, which seems a bit low. Going to try it against my own data...
Screen Shot 2023-10-28 at 10 09 03 PM

shankari:
While testing against my own data, ran into the following error

```
Traceback (most recent call last):
  File "/usr/src/app/pages/segment_trip_time.py", line 181, in generate_content_on_endpoints_change
    mode_by_section_id = db_utils.query_inferred_sections_modes(
  File "/usr/src/app/utils/db_utils.py", line 247, in query_inferred_sections_modes
    return esds.cleaned2inferred_section_list(sections)
  File "/usr/src/app/emission/storage/decorations/section_queries.py", line 51, in cleaned2inferred_section_list
    matching_inferred_section = cleaned2inferred_section(section_userid.get('user_id'), section_userid.get('section'))
  File "/usr/src/app/emission/storage/decorations/section_queries.py", line 45, in cleaned2inferred_section
    curr_predicted_entry = _get_inference_entry_for_section(user_id, section_id, "analysis/inferred_section", "data.cleaned_section")
  File "/usr/src/app/emission/storage/decorations/section_queries.py", line 66, in _get_inference_entry_for_section
    assert len(ret_list) <= 1, "Found len(ret_list) = %d, expected <=1" % len(ret_list)
AssertionError: Found len(ret_list) = 807, expected <=1
```

shankari:
Looking at the logs, we have

```
op-admin-dash-dashboard-1  | DEBUG:root:About to query {'metadata.key': 'analysis/inferred_section', 'user_id': UUID('9c084ef4-2f97-4196-bd37-950c17938ec6'), 'data.cleaned_section': ObjectId('643874ce88f9b4eda2beca67')}
op-admin-dash-dashboard-1  | DEBUG:root:About to query {'metadata.key': 'analysis/inferred_section', 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'data.cleaned_section': ObjectId('644e58ecb14cecd84298aae4')}
op-admin-dash-dashboard-1  | DEBUG:root:Found no inferred prediction, returning None
op-admin-dash-dashboard-1  | DEBUG:root:About to query {'metadata.key': 'analysis/inferred_section', 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'data.cleaned_section': ObjectId('64532ffeedc48e75d0268b04')}
op-admin-dash-dashboard-1  | DEBUG:root:Found no inferred prediction, returning None
op-admin-dash-dashboard-1  | DEBUG:root:About to query {'metadata.key': 'analysis/inferred_section', 'user_id': UUID('16c2d3cd-6d62-42dc-98df-6d927cd9a3c8'), 'data.cleaned_section': ObjectId('62db2032a6977e4c0214befe')}
op-admin-dash-dashboard-1  | DEBUG:root:About to query {'metadata.key': 'analysis/inferred_section', 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'data.cleaned_section': ObjectId('644df8edea199f1d0473e301')}
```

so it must be one of these sections

shankari:
Bingo!

```
# ./e-mission-py.bash
Python 3.9.18 | packaged by conda-forge | (main, Aug 30 2023, 03:49:32)
[GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import emission.core.get_database as edb
Connecting to database URL mongodb://db/openpath_stage
>>> from uuid import UUID
>>> from bson.objectid import ObjectId
>>> edb.get_analysis_timeseries_db().find({'metadata.key': 'analysis/inferred_section', 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'data.cleaned_section': ObjectId('644df8edea199f1d0473e301')})
<pymongo.cursor.Cursor object at 0x7f1e66313970>
>>> edb.get_analysis_timeseries_db().count_documents({'metadata.key': 'analysis/inferred_section', 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'data.cleaned_section': ObjectId('644df8edea199f1d0473e301')})
807
>>> edb.get_analysis_timeseries_db().count_documents({'metadata.key': 'analysis/inferred_section', 'user_id': UUID('16c2d3cd-6d62-42dc-98df-6d927cd9a3c8'), 'data.cleaned_section': ObjectId('62db2032a6977e4c0214befe')})
1
```

This is almost certainly due to 
https://github.com/e-mission/e-mission-docs/issues/927#issuecomment-1606541814

My current guess is that it might be due to multiple calls to overpass failing, but in that previous issue, I see:
> That seemed weird, since there did not appear to be any errors while generating the mode inference.

shankari:
It's really weird that there are still exactly 807 matching entries. Maybe we can spend a little time today to investigate (at least on the side)

shankari:
After resetting the pipeline for that user, we get 30 trips. Which is not that much but is at least greater than 3. 
I am going to merge this for now, but also poke around a bit to see if this is actually correct.
Screen Shot 2023-10-29 at 9 19 00 AM

pages/segment_trip_time.py
```diff
@@ -0,0 +1,284 @@
...
+        )
+        df['mode'] = df['section'].apply(
+            lambda section_id: mode_by_section_id[str(section_id)].name
+        )
+        median_trip_time = df['duration'].median()
+        times = pd.to_datetime(df['start_fmt_time'], errors='coerce', utc=True)
+        duration_per_hour = format_duration_df(
+            df.groupby(times.dt.hour).agg({'duration': 'median', 'section': 'count'}),
+            time_column_name='Hour',
+        )
```
shankari, line 192: I just tried to use this myself, and having the hours in UTC is very annoying. We already have the split out components **in local time** in the `data.start_local_dt` field

From `emission/core/wrapper/section.py`

```
             "start_local_dt": ecwb.WrapperBase.Access.WORM, # ********searchable datatime in local time of start location
```

I will spend ~ 10 mins trying to fix this myself while merging the change, but will file a cleanup issue if I can't get that to work.

shankari:
After resetting and re-running the pipeline, we get 14k points at the start and 824 points at the end.
But only 31 overlaps. I bet it is because of shorter segments. Let's try with a smaller segment of road so that it is more likely to be in the same section.

```
op-admin-dash-dashboard-1  | DEBUG:root:Found 14390 results
op-admin-dash-dashboard-1  | DEBUG:root:After de-duping, converted 14390 points to 14390

op-admin-dash-dashboard-1  | DEBUG:root:Found 824 results
op-admin-dash-dashboard-1  | DEBUG:root:After de-duping, converted 824 points to 824
```

shankari:
As expected, there are more trips for a short segment although still not as many as we would like.
Screen Shot 2023-10-29 at 9 48 01 AM

shankari:
Ran into the duplicate entries for another user
https://github.com/e-mission/e-mission-docs/issues/927#issuecomment-1784168435

We might want to write a check for this and run it on production before pushing it out.

Let's switch to open access for a bit to see if things are better.

achasmita:
<details>
<summary> I got more trips when I selected bigger zone: </summary>
Screen Shot 2023-11-08 at 3 35 44 PM
Screen Shot 2023-11-08 at 3 36 00 PM
Screen Shot 2023-11-08 at 3 36 14 PM
</details>

achasmita:
> I've notices that considering a bigger zone results in more trips. However, I am still unclear about determining appropriate zone size, especially given the presence of different modes. As the instruction metion that "zone length should somewhat match the distance a vehicle can cross at the maximum allowed speed in 30 seconds".
Screen Shot 2023-11-15 at 9 53 56 PM
Screen Shot 2023-11-15 at 9 54 12 PM

shankari:
@achasmita I don't see how big these zones are. How did you pick them?
How do you validate that the number of trips is correct?

achasmita:
> @achasmita I don't see how big these zones are. How did you pick them? How do you validate that the number of trips is correct?

<details>

<summary> While selecting zone, I checked trip table to find the best areas based on latitude and longitude that had the most trips: </summary>

Screen Shot 2023-11-18 at 9 02 05 PM

Screen Shot 2023-11-18 at 9 21 20 PM

Screen Shot 2023-11-18 at 9 21 35 PM

> I also observed locs_matching_start  amd locs_matching_end data before removing duplicates and after removing duplicates and once they are filtered: 

>> locs_matching_start before and after removing duplicates (left) and locs_matching_end before and after removing duplicates (right)
image

- After selecting the start and end zone, duplicate sections are removed. It is keeping first occured section and removing the other occurences. 
>>data after merged and filtered (After removing duplicates and after filtering data for which  [merged['idx_x'] < merged['idx_y']])
image
After observing data what I can see now is:

-  Also if there is an overlap it will not include those section as both section will have same idx.
- And manually verified this result:
Screen Shot 2023-11-19 at 5 06 40 PM


</details>

achasmita:
Also I plotted some co-ordinates (25-30) from trip table  manually by checking it on google map and it was giving the coorrect result.

achasmita:
I tried exploring start and end zone with different segment size:
-I was getting very few trips
- I also explored data in both zone(start/end)
- I was not able to figure out anything new.

shankari:
@achasmita Thank you for adding additional examples with the results of your investigation. I have some more questions.
1. You said that you picked ~ 25-30 trips, but there are actually 105 trips displayed above. Can you explain this discrepancy?
2. I don't understand the follow-on.
> locs_matching_start before and after removing duplicates (left) and locs_matching_end before and after removing duplicates (right)
   > After selecting the start and end zone, duplicate sections are removed. It is keeping first occured section and removing the other occurences.
   But we first select the start and end zones - that is what allows us to find the `locs_matching_start` and `locs_matching_end`, right?

Can you also expand on what you did in "I also explored data in both zone(start/end)"? How did you explore the data, and what were the results?

achasmita:
> @achasmita Thank you for adding additional examples with the results of your investigation. I have some more questions.
> 
> 1. You said that you picked ~ 25-30 trips, but there are actually 105 trips displayed above. Can you explain this discrepancy?
> 2. I don't understand the follow-on.
> 
> > locs_matching_start before and after removing duplicates (left) and locs_matching_end before and after removing duplicates (right)
> > After selecting the start and end zone, duplicate sections are removed. It is keeping first occured section and removing the other occurences.
> > But we first select the start and end zones - that is what allows us to find the `locs_matching_start` and `locs_matching_end`, right?
> 
> Can you also expand on what you did in "I also explored data in both zone(start/end)"? How did you explore the data, and what were the results?

The above screeshot was just to make sure if i am selecting correct size of zone, I will find the other screenshot for 25-30 trips and post it soon.

For data, I printed tops 50 and bottom 50 data after removing duplicates and compared it with data in trip table to see if i can figure out if any data is missing.

shankari:
@achasmita 

> For data, I printed tops 50 and bottom 50 data after removing duplicates and compared it with data in trip table to see if i can figure out if any data is missing.

Can you expand on this?

achasmita:
> @achasmita
> 
> > For data, I printed tops 50 and bottom 50 data after removing duplicates and compared it with data in trip table to see if i can figure out if any data is missing.
> 
> Can you expand on this?

- I printed data on locs_matching_start once the duplicates are excluded
- I compared the result with data on trip table
- As I selected start zone considering start coordinates (data.start_loc_coordinates) in trip table.

shankari:
My main concern with this is we were getting very few trips displayed for basically any start/end combo.
- I want to see "large number" of trips OR
- I want some explanation of why they don't exist

@achasmita was not able to get (1) without making the polygons really large and wasn't able to come up with (2).
The next steps that I was going to do was to look at locations where I anticipate having a lot of trips and then seeing why they don't show up in the list

Aa a concrete example, on staging, I would expect to see a lot of trips from my house to the library or to the grocery store nearby or to my kids' school. In particular, I would expect to see at least 100 trips from my house to the local school. Similarly, in the Denver area, you could see the locations that are hotspots in the heatmap and try to see if there are trips between them.

JGreenlee:
While trying this branch, I initially got this error:

```
AttributeError: module 'emission.storage.decorations.section_queries' has no attribute 'cleaned2inferred_section_list'
```

I realized this is because this branch was using an old image of `e-mission-server` (shankari/e-mission-server:gis-based-mode-detection_2023-04-21--54-09).
Which was probably before `cleaned2inferred_section_list` was added.

Updated to the most recent (shankari/e-mission-server:master_2024-02-10--19-38) and rebuilt.

<hr>

Now there's a different error:

```
Traceback (most recent call last):
  File "/usr/src/app/pages/segment_trip_time.py", line 181, in generate_content_on_endpoints_change
    mode_by_section_id = db_utils.query_inferred_sections_modes(
  File "/usr/src/app/utils/db_utils.py", line 200, in query_inferred_sections_modes
    return esds.cleaned2inferred_section_list(sections)
  File "/usr/src/app/emission/storage/decorations/section_queries.py", line 51, in cleaned2inferred_section_list
    matching_inferred_section = cleaned2inferred_section(section_userid.get('user_id'), section_userid.get('section'))
  File "/usr/src/app/emission/storage/decorations/section_queries.py", line 45, in cleaned2inferred_section
    curr_predicted_entry = _get_inference_entry_for_section(user_id, section_id, "analysis/inferred_section", "data.cleaned_section")
  File "/usr/src/app/emission/storage/decorations/section_queries.py", line 66, in _get_inference_entry_for_section
    assert len(ret_list) <= 1, "Found len(ret_list) = %d, expected <=1" % len(ret_list)
AssertionError: Found len(ret_list) = 807, expected <=1
```

There are 807 inferred sections for one cleaned section??

<hr>

I inspected the logs to see which UUID + section this is happening for. It happens on:
`DEBUG:root:About to query {'metadata.key': 'analysis/inferred_section', 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'data.cleaned_section': ObjectId('644df8edea199f1d0473e301')}`

Sure enough, there are 807 inferred section entries for that UUID and that cleaned section.
They must be duplicate entries because they are identical except for their `_id`
```python
query = {
  'metadata.key': 'analysis/inferred_section',
  'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'),
  'data.cleaned_section': ObjectId('644df8edea199f1d0473e301')
}

r = edb.get_analysis_timeseries_db().find(query)
for i in r:
    print(i)
```

> ```{'_id': ObjectId('644e0e92ea199f1d04742709'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644e6f67b14cecd84298f6b2'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644e7949cdabcb78bc676484'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644e88388464a359f04a7c74'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644e95d6fac3c75f1a08eb28'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644ea4395e9649cf426dd33d'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644eb16ab2270c7ba1ae53fc'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644ebf8b7c7a20f08cd4d7dc'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644ecddcb4d41475b8330bee'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644edbdf3e1640bffa2a6051'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644ee98a98049428510657bb'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644ef78671732fce1ba61e81'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644f05e5635c15953692eeb0'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> {'_id': ObjectId('644f13cf3a381d0218807744'), 'user_id': UUID('d83a43a1-df6b-42ed-986f-f5b5f6150221'), 'metadata': {'key': 'analysis/inferred_section', 'platform': 'server', 'write_ts': 1682831597.6901028, 'time_zone': 'America/Los_Angeles', 'write_local_dt': {'year': 2023, 'month': 4, 'day': 29, 'hour': 22, 'minute': 13, 'second': 17, 'weekday': 5, 'timezone': 'America/Los_Angeles'}, 'write_fmt_time': '2023-04-29T22:13:17.690103-07:00'}, 'data': {'source': 'SmoothedHighConfidenceMotion', 'trip_id': ObjectId('644df8dbea199f1d0473e2ff'), 'start_ts': 1659712158.517581, 'start_local_dt': {'year': 2022, 'month': 8, 'day': 5, 'hour': 8, 'minute': 9, 'second': 18, 'weekday': 4, 'timezone': 'America/Los_Angeles'}, 'start_fmt_time': '2022-08-05T08:09:18.517581-07:00', ...
> ...
> ```

I removed those duplicates and tried again, but there appears to be duplicates for all the other sections too.

I am unsure why there are so many duplicates.
Do they exist in the dataset? Or did I create duplicates while I was loading in the dataset?

JGreenlee:
I removed the duplicates manually with this script:

```python
import emission.core.get_database as edb

# get cleaned sections
cleaned = edb.get_analysis_timeseries_db().find({
    'metadata.key': 'analysis/cleaned_section',
})

cleaned_ct = 0
for c in cleaned:
    cleaned_ct += 1
    query_inferred = {
      'metadata.key': 'analysis/inferred_section',
      'user_id': c['user_id'],
      'data.cleaned_section': c['_id']
    }
    
    first_inferred = edb.get_analysis_timeseries_db().find_one(query_inferred)
    if first_inferred is None:
        print(f"cleaned section {cleaned_ct} had no inferred sections")
        continue

    # remove all of those entries unless the ID is the first inferred section
    dedup_query = {
      'metadata.key': 'analysis/inferred_section',
      'user_id': c['user_id'],
      'data.cleaned_section': c['_id'],
      '_id': {'$ne': first_inferred['_id']}
    }
    delete_result = edb.get_analysis_timeseries_db().delete_many(dedup_query)
    print(f"removed {delete_result.deleted_count} duplicates from cleaned section {cleaned_ct}")
```

It took awhile to run.

<hr>

Now I am finally able to test the Segment Trip Time page.

> I would expect to see a lot of trips from my house to the library or to the grocery store nearby or to my kids' school. In particular, I would expect to see at least 100 trips from my house to the local school.

For trips from home to school, I found 222 trips spanning from July 2022 to December 2023. This seems to align with expectations because there are about 180 school days in 1 year.

The boxes I used were about the size of 1 block. I will follow up with smaller boxes

JGreenlee:
### Home to school
 
\<image>
Home to school with boxes about the size of 1 block, got 222 trips
 
\<image>
Home to school with boxes about half that size, 197 trips (still pretty good)

The trips span from July 2022 to December 2023. This seems to align with expectations because there are about 180 school days in 1 year.
Based on the dates, it generally makes sense – large gaps are observed for summer break etc
\<image>

So I do think this is probably a pretty comprehensive measure of this repeated trip
 
 
### Home to viola class

\<image>
Home to viola class, using smaller boxes, 52 trips

\<image>
This trip is observed almost every week, sometimes twice in the same day but not always

\<image>
1 "bicycling" trip to viola class which took 18 minutes - I think this was probably just mislabeled


### Methodology for drawing boxes
 
I found these usage guidelines quite helpful and accurate:
\<image>

For reference, below are the heatmaps around those 3 areas of interest. I drew the boxes considering where the locus of activity seems to be for each area (and considering the guidelines above)
 
(I wonder if this tool could potentially be even more useful + easy to use if a heatmap was actually overlayed on the start/end selection area? I found myself switching back and forth often)

\<image>
\<image>
\<image>


### Conclusion

Based on this, the tool does appear to work as expected. It captured a fairly comprehensive, if not fully comprehensive, picture of the above recurring trips. I also briefly validated the tool against my own travel data.
I think the instructions for usage are clear as well.

The only changes I might suggest would include a heatmap overlay to make it easier to identify places of activity while drawing boxes, and potentially a toggle to "swap" start and end locations. (I see a common use case where the user has seen the duration from A to B - now they want to see the duration from B to A, but they don't want to have to re-draw the boxes)

shankari:
@JGreenlee thanks for the comprehensive review! Given the length of time that this has been pending, I will merge the changes now for the next round and we can address the UX improvements in a subsequent round.

JGreenlee:
@shankari Great!
When this is merged, the `Dockerfile` must be updated with a more recent image of `e-mission-server`

> While trying this branch, I initially got this error:
> 
> ```
> AttributeError: module 'emission.storage.decorations.section_queries' has no attribute 'cleaned2inferred_section_list'
> ```
> 
> I realized this is because this branch was using an old image of `e-mission-server` (shankari/e-mission-server:gis-based-mode-detection_2023-04-21--54-09). Which was probably before `cleaned2inferred_section_list` was added.
> 
> Updated to the most recent (shankari/e-mission-server:master_2024-02-10--19-38) and rebuilt.

JGreenlee:
Actually it looks like you just did that a few days ago!

JGreenlee:
I resolved the merge conflicts for this feature and updated it to observe the global filters (which were added since this feature was created).

> ### apply date range & uuid filters to the 'segment trip time' page
> The 'segment trip time' page was written before we had the global filters implemented. So we need to patch them in now to get this feature up to speed and merged.
> 
> Datepicker values and excluded uuids (if any) are passed through to the query_segments_crossing_endpoints function. Within this function is the DB call. We pass a time_query arg based on the selected dates + timezone. We also pass a query in the extra_query_list arg that will exclude any entries where the user_id is found in the excluded_uuids list.

@shankari 
These changes are on my branch https://github.com/JGreenlee/op-admin-dashboard/tree/segment_trip_time_resolved_conflicts

PR #63, opened 2023-08-31
Updated readme and added demographics table

achasmita:


docker-compose-dev.yml
```diff
@@ -12,7 +12,8 @@ services:
       DASH_DEBUG_MODE: "True"
       DASH_SILENCE_ROUTES_LOGGING: "False"
       DASH_SERVER_PORT: 8050
-      DB_HOST: db
+      # DB_HOST: db
+      DB_HOST: "mongodb://db/openpath_prod_durham"
```
shankari, line 16: This should not be hardcoded since the DB will change depending on the dataset used for testing. Please revert.

utils/constants.py
```diff
@@ -44,3 +44,82 @@
...
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.attr.xmlns:jr',
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.attr.xmlns:orx',
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.group_yk8eb99.At_your_primary_job_do_you_ha',
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.group_yk8eb99.Which_best_describes_your_prim',
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.group_yk8eb99.Do_you_work_full_time_or_part_',
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.group_yk8eb99.Do_you_have_the_option_of_work',
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.group_yk8eb99.Please_describe_your_primary_job',
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.group_yk8eb99.Do_you_have_more_than_one_job',
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.group_yk8eb99.What_days_of_the_week_do_you_t',
+    'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.group_yk8eb99.How_many_days_do_you_usually_w_001',
```
shankari, line 92: We cannot hardcode these because the survey can be different for each partner.
You should exclude the entries that the config wants to be excluded (currently marked as TODO) and potentially exclude several of the entries here. But we should not have a list of valid cols

utils/db_utils.py
```diff
@@ -98,6 +100,31 @@ def query_confirmed_trips(start_date, end_date):
     # logging.debug("After filtering, the actual data is %s" % df.head().trip_start_time_str)
     return df
 
+def query_demographics(start_date, end_date):
+    start_ts, end_ts = None, datetime.max.timestamp()
+    if start_date is not None:
+        start_ts = datetime.combine(start_date, datetime.min.time()).timestamp()
+
+    if end_date is not None:
+        end_ts = datetime.combine(end_date, datetime.max.time()).timestamp()
```
shankari, line 109: demographic data is not the same as trip data - it is typically only answered once, at the beginning of the program. 
Trips come in throughout the program. We cannot blindly apply the trip filters to the demographic data

utils/permissions.py
```diff
@@ -88,5 +88,11 @@ def get_uuids_columns():
     return columns
 
 
+def get_demographics_columns():
+    columns = set(constants.VALID_DEMOGRAPHICS_COLS)
```
shankari, line 92: we should remove this (see note above)

pages/data.py
```diff
@@ -58,6 +59,11 @@ def render_content(tab, store_uuids, store_trips):
             col['label'] for col in perm_utils.get_allowed_named_trip_columns()
         )
         has_perm = perm_utils.has_permission('data_trips')
+    elif tab == 'tab-demographics-datatable':
+        data = store_demographics["data"]
+        columns = perm_utils.get_demographics_columns() # TODO
```
shankari, line 64: is this still TODO?

app_sidebar_collapsible.py
```diff
@@ -204,6 +205,25 @@ def update_store_trips(start_date, end_date):
     return store
 
 
+@app.callback(
+    Output("store-demographics", "data"),
+    Input('date-picker', 'start_date'),
+    Input('date-picker', 'end_date'),
```
shankari, line 211: note comment above: we don't need this

docker/start.sh
```diff
@@ -7,7 +7,7 @@ if [ -z ${DB_HOST} ] ; then
     local_host=`hostname -i`
     sed "s_localhost_${local_host}_" conf/storage/db.conf.sample > conf/storage/db.conf
 else
-    sed "s_localhost_${DB_HOST}_" conf/storage/db.conf.sample > conf/storage/db.conf
+    sed "s-localhost-${DB_HOST}-" conf/storage/db.conf.sample > conf/storage/db.conf
```
shankari, line 10: This is specific to the DB format that you are testing against and should not be checked in.

shankari (changes requested):

utils/db_utils.py
```diff
@@ -98,6 +100,22 @@ def query_confirmed_trips(start_date, end_date):
...
 
+def query_demographics():
+    ts = esta.TimeSeries.get_aggregate_time_series()
+    
+    entries = ts.find_entries(
+        key_list=["manual/demographic_survey"],
+    )
+
+    df = pd.json_normalize(list(entries))
+
```
shankari, line 111: I am not going to hold this up but have logged a potential improvement for later.

docker-compose-dev.yml
```diff
@@ -47,3 +47,4 @@ networks:
 
 volumes:
   mongo-data:
+
```
shankari, line 50: extraneous whitespace, revert. You could just revert the entire file

pages/data.py
```diff
@@ -58,6 +59,11 @@ def render_content(tab, store_uuids, store_trips):
             col['label'] for col in perm_utils.get_allowed_named_trip_columns()
         )
         has_perm = perm_utils.has_permission('data_trips')
+    elif tab == 'tab-demographics-datatable':
+        data = store_demographics["data"]
+        columns = list(data[0].keys())
+        has_perm = perm_utils.has_permission('data_demographics')
```
shankari, line 65: why are we not supporting excluded columns here? Please see my previous comment
https://github.com/e-mission/op-admin-dashboard/pull/63#discussion_r1313033861

>  You should exclude the entries that the config wants to be excluded (currently marked as TODO) and potentially exclude several of the entries here. But we should not have a list of valid cols

I don't see where you are excluding the entries that are marked as needing exclusion in the config.

shankari (changes requested):
Almost there!

pages/data.py
shankari: For backwards compatibility, if the config doesn't have the `data_demographic_columns_exclude`, then you can set it to a blank array. add a note to remove the backwards compat change after a couple of years when all the current studies are done.

docker-compose-dev.yml
```diff
@@ -46,4 +46,4 @@ networks:
   emission:
 
 volumes:
-  mongo-data:
+  mongo-data:
```
shankari, line 49: Extraneous whitespace change, please revert the entire file.

README.md
```diff
@@ -65,6 +65,7 @@ These are all the permissions that you can specify:
 ### Data Page
 - `data_uuids`: User can view the UUIDs data in the Data page.
 - `data_trips`: User can view the trips data in the Data page.
+- `data_demographics`: User can view the demographics data in the Data page.
```
shankari, line 68: you should add an entry for `data_demographics_columns_exclude` now that you have added that config option.

pages/data.py
```diff
@@ -58,6 +59,13 @@ def render_content(tab, store_uuids, store_trips):
             col['label'] for col in perm_utils.get_allowed_named_trip_columns()
         )
         has_perm = perm_utils.has_permission('data_trips')
+    elif tab == 'tab-demographics-datatable':
+        data = store_demographics["data"]
+        columns = list(data[0].keys())
+        for column in perm_utils.permissions.get("data_demographics_columns_exclude", []):
+            columns.discard(column)
```
shankari, line 66: **(future)** Is there a reason that you discard here (in `data.py`) while the other calls to `discard` are in `permissions.py`?
https://github.com/search?q=repo%3Ae-mission%2Fop-admin-dashboard%20discard&type=code

shankari (changes requested):
Almost there.
Just some documentation/formatting cleanup and one clarification request...

Please also indicate "Testing done".
"If it is not tested, it is broken"

utils/db_utils.py
shankari: Potential improvement logged here:
https://github.com/e-mission/op-admin-dashboard/issues/51#issuecomment-1709275660

shankari:
@achasmita 

Code looks fine.
Please indicate "Testing done" before I can merge.
"If it is not tested, it is broken"

achasmita:
> @achasmita
> 
> Code looks fine. Please indicate "Testing done" before I can merge. "If it is not tested, it is broken"
Screen Shot 2023-09-12 at 11 15 57 AMYes, testing is done and table looks fine.

shankari:
@achasmita thank you for including the screenshot.
Please do that for all PRs going forward.

However, the testing is not sufficient because it doesn't actually show any of the demographic survey fields.
How do I know if the demographic survey is actually being parsed properly?

Also, please automatically exclude the vast majority of the metadata fields. We don't need the `metadata.key` (which is the same for all entries, or the `metadata.type`, ...

Please updated with more descriptive screenshots that show what you are actually testing.

achasmita:
> @achasmita thank you for including the screenshot. Please do that for all PRs going forward.
> 
> However, the testing is not sufficient because it doesn't actually show any of the demographic survey fields. How do I know if the demographic survey is actually being parsed properly?
> 
> Also, please automatically exclude the vast majority of the metadata fields. We don't need the `metadata.key` (which is the same for all entries, or the `metadata.type`, ...
> 
> Please updated with more descriptive screenshots that show what you are actually testing.

Here is the one after removing metadata.key and including demographic survey fields.Screen Shot 2023-09-12 at 3 32 45 PM

shankari:
Does this only remove `metadata.key`? Please note my comment

> Also, please automatically exclude the vast majority of the metadata fields.

achasmita:
> Does this only remove `metadata.key`? Please note my comment
> 
> > Also, please automatically exclude the vast majority of the metadata fields.

i mean all the columns that starts with metadata.

achasmita:
Here is the screenshot of all the columns it is displaying now:

Screen Shot 2023-09-12 at 3 41 06 PM
Screen Shot 2023-09-12 at 3 42 57 PM
Screen Shot 2023-09-12 at 3 43 22 PM


Screen Shot 2023-09-12 at 3 42 31 PM
Screen Shot 2023-09-12 at 3 43 36 PM

Screen Shot 2023-09-12 at 3 42 06 PM

shankari:
Please push these changes and I can merge

utils/db_utils.py
```diff
@@ -106,10 +106,13 @@ def query_demographics():
     if not df.empty:
         for col in constants.BINARY_DEMOGRAPHICS_COLS:
             if col in df.columns:
-                df[col] = df[col].apply(str)  
-    df.drop(columns=['data.xmlResponse', 'data.name', 'data.version', 'data.label'], inplace=True) 
+                df[col] = df[col].apply(str) 
+    columns_to_drop = [col for col in df.columns if col.startswith("metadata")]
+    df.drop(columns= columns_to_drop, inplace=True) 
+    df.drop(columns=['data.xmlResponse', 'data.name', 'data.version', 'data.label', 'data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.attr.id','data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.attr.xmlns:jr','data.jsonDocResponse.aSfdnWs9LE6q8YEF7u9n85.attr.xmlns:orx'], inplace=True)
```
shankari, line 112: I don't think we should hardcode any of the form ids (such as `aSfdnWs9LE6q8YEF7u9n85`) because I believe they can be different from form to form. But the other changes are fine.

shankari: e.g. the form ID for the time use survey, for example, is `time_use_survey_form_v9_1`
https://github.com/sebastianbarry/nrel-openpath-deploy-configs/blob/surveys-info-and-surveys-data/survey-resources/data-json/time-use-survey-form-v9.json

shankari (approved):
I am going to approve this for now and we can tweak based on feedback.

achasmita:
new screenshots ![Screen Shot 2023-09-12 at 4 26 11 PM
Screen Shot 2023-09-12 at 4 25 50 PM](https://github.com/e-mission/op-admin-dashboard/assets/79387860/6d40745c-c61b-437f-81af-35398fa57eb0)

Issue #55, opened 2023-06-13
Support O-D and trajectory downloads


shankari:
We currently have trip tables available for download.
The trip tables contain user, distance, duration, time and labels in the common (anonymized) case.
They can contain O-D locations as well, but most current deployers have chosen not to include them.
Instead the O-D information is displayed in maps.

But we now have a request to support download of O-D and trajectory information for planning, *without* the connection to the user. So this would involve supporting downloads of tables with the O-D, time, trajectory, and labels but without the user id.
We need to create those tables so that they can be downloaded for better planning and analysis.

shankari:
Based on the target audience for this feature, I am going to assume that a table will be preferable - using excel with the data is the lowest common denonimator.

@achasmita you should be able to read the `analysis/recreated_location` and display it as a table in the same way that the trip table (`analysis/confirmed_trip`) or the demographic table (`manual/demographic_survey`) are displayed. It can go on the same page, in a new tab.

Once you are done with that basic export, we might want to support other options:
1. Toggle between the cleaned locations (`analysis/recreated_location`) and the raw locations (`background/location`)
2. Provide an option to download timestamped geojson instead https://python-visualization.github.io/folium/latest/user_guide/plugins/timestamped_geojson.html

Note that, like in the trip table, we want to determine whether or not to exclude the user_id and the section_id based on a config option.

PR #66, opened 2023-09-14
Added trajectory table to data page

achasmita:


shankari (changes requested):
I think that this is generally fine, which is not surprising since it is essentially a copy of the demographic changes.

However:
- since we have already missed the previous release train, *and*
- the trajectory data is many orders of magnitude larger than the trip data, so loading it is likely to be slow

While we can and will address this in a more principled fashion in
https://github.com/e-mission/op-admin-dashboard/issues/55
I also don't want to run the risk of a regression by making the admin dashboard unusable with this change.

So, I think we should, at least temporarily, change the loading pattern for this table. Instead of loading the data when the admin dashboard is launched, we should load the data only when the user accesses the "trajectories" tab.

That will keep the admin dashboard performant overall, and reserve the performance hit for if/when the admin actually clicks on the trajectory tab.

achasmita:
Here is the image for all the columns in the table.
Screen Shot 2023-09-19 at 8 02 59 PM

Screen Shot 2023-09-19 at 8 03 56 PM

Screen Shot 2023-09-19 at 8 04 40 PM

shankari:
@achasmita is this ready for review? It is still in draft mode...

pages/data.py
```diff
@@ -63,6 +76,12 @@ def render_content(tab, store_uuids, store_trips, store_demographics):
         data = store_demographics["data"]
         columns = list(data[0].keys())
         has_perm = perm_utils.has_permission('data_demographics')
+    elif tab == 'tab-trajectories-datatable':
+        if store_trajectories == {}:
+            store_trajectories = update_store_trajectories()
```
shankari, line 81: I am glad that you have this in a separate commit - it makes understanding the code easier. I think that this would also benefit from a short code comment on why it is different from the other tabs so that people have a rough sense without having to look at the blame immediately.

shankari (changes requested):
This is a simple change so this is fine for now. But gentle reminder to add additional context to the issue as you explore increasingly complex implementations - e.g. when you read up on how to load the trajectories on demand, where did you find that from? How do you know that it works - e.g. how do you know that it *doesn't* load at startup?

Only requested change is a new code comment.

You also need to mark it as "non-draft" and resolve the conflict before I merge

shankari:
I am merging this for now, but I was looking for more detailed comments, on the lines of:

> when the user clicks any of the tabs, this function is invoked. The data for the other tabs is loaded at start.
> but for this tab, we load the data if it has not yet been loaded.
> however, if the data has already been loaded, we don't re-load
> if the data changes while the dashboard is open..... (you need to fill this in).

Issue #65, opened 2023-09-13
Demographic table cleanup


achasmita:
Need to remove extraneous columns from demographic table that are not necessary.

shankari:
After the demographic table was added, a couple of the programs had their admin dashboard fail to startup.
The error is 

```
Traceback (most recent call last):
File "/usr/src/app/app_sidebar_collapsible.py", line 169, in <module>
demographics_data = update_store_demographics()
File "/usr/src/app/app_sidebar_collapsible.py", line 160, in update_store_demographics
df = query_demographics()
File "/usr/src/app/utils/db_utils.py", line 116, in query_demographics
df.drop(columns=['xmlns:jr', 'xmlns:orx', 'id'], inplace = True)
File "/root/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/util/_decorators.py", line 331, in wrapper
return func(*args, **kwargs)
File "/root/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/frame.py", line 5399, in drop
return super().drop(
File "/root/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/util/_decorators.py", line 331, in wrapper
return func(*args, **kwargs)
File "/root/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/generic.py", line 4505, in drop
obj = obj._drop_axis(labels, axis, level=level, errors=errors)
File "/root/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/generic.py", line 4546, in _drop_axis
new_axis = axis.drop(labels, errors=errors)
File "/root/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 6934, in drop
raise KeyError(f"{list(labels[mask])} not found in axis")
KeyError: "['xmlns:jr', 'xmlns:orx', 'id'] not found in axis"
```

@achasmita

achasmita:
> After the demographic table was added, a couple of the programs had their admin dashboard fail to startup. The error is
> 
> ```
> Traceback (most recent call last):
> File "/usr/src/app/app_sidebar_collapsible.py", line 169, in <module>
> demographics_data = update_store_demographics()
> File "/usr/src/app/app_sidebar_collapsible.py", line 160, in update_store_demographics
> df = query_demographics()
[...]

PR #67, opened 2023-09-14
Removing extraneous columns from demographic table

achasmita:


shankari:
@achasmita
1. please indicate testing done. "If it is not tested, it is broken"
2. From a clean commit history perspective, the second commit message is a duplicate of the first and is incorrect
3. how come these didn't show up in the testing of the previous PR (#63)? Are these columns not in the test dataset that has been shared with you?

utils/db_utils.py
```diff
@@ -107,13 +107,14 @@ def query_demographics():
...
-    df.columns=[col.rsplit('.',1)[-1] if col.startswith('data.jsonDocResponse.') else col for col in df.columns]  
-    df.drop(columns=['xmlns:jr', 'xmlns:orx', 'id'], inplace = True) 
+        columns_to_drop = [col for col in df.columns if col.startswith("metadata")]
+        df.drop(columns= columns_to_drop, inplace=True) 
+        modified_columns = perm_utils.get_demographic_columns(df.columns)  
+        df.columns = modified_columns 
+        df.columns=[col.rsplit('.',1)[-1] if col.startswith('data.jsonDocResponse.') else col for col in df.columns]  
+        for col in constants.EXCLUDED_DEMOGRAPHICS_COLS:
+            if col in df.columns:
+                df.drop(columns= [col], inplace=True)
```
shankari, line 117: why isn't this just 

```suggestion

        df.drop(columns=constants.EXCLUDED_DEMOGRAPHICS_COLS, inplace=True)
```

achasmita:
> @achasmita
> 
> 1. please indicate testing done. "If it is not tested, it is broken"
> 2. From a clean commit history perspective, the second commit message is a duplicate of the first and is incorrect
> 3. how come these didn't show up in the testing of the previous PR ([Updated readme and added demographics table #63](https://github.com/e-mission/op-admin-dashboard/pull/63))? Are these columns not in the test dataset that has been shared with you?

No, those columns are not available in the dataset provided to me. So it was all good when I tested locally. But I saw extraneous columns while in staging.

utils/db_utils.py
achasmita: As this columns are not i the data set I am working, we cannot directly exclude all those columns.

shankari:
Merging this now to try and resolve showstopper
https://github.com/e-mission/op-admin-dashboard/issues/65

shankari:
@achasmita I am squash-merging this, please take it into account while pulling from upstream.

PR #69, opened 2023-09-27
Used find_entries_count() function to count database entries.

achasmita:


shankari:
@achasmita I rebased the PR a couple of times, we now see only the new commit.
**Please indicate "testing done"**
it would be better to actually make sure to indicate testing done for each commit

achasmita:
> @achasmita I rebased the PR a couple of times, we now see only the new commit. **Please indicate "testing done"** it would be better to actually make sure to indicate testing done for each commit

I will make sure from next time. Yes, it is working fine.

shankari:
@achasmita screenshots plz

achasmita:
> @achasmita screenshots plz

I don't have UUID table so the count for active user is 0 both before and after the test while confirmed trip is 6829 both before and after the test.

achasmita:
Screen Shot 2023-09-27 at 4 59 40 PM

shankari:
@achasmita is this before or after? I would like to see

#### Before screenshot
...

#### After screenshot
...

achasmita:
The screenshot is same for both before and after.

shankari:
Right, so you should say that, or put two screenshots for before and after

achasmita:
ok here is the one before:
Screen Shot 2023-09-27 at 4 59 40 PM

And after changes

Screen Shot 2023-09-27 at 5 11 46 PM

shankari:
I do note that the timestamps on them are different, indicating that you took them at different times 😄

Issue #70, opened 2023-09-28
Trajectory table not visible in staging.


achasmita:
It does work on dev against the durham dataset.

shankari:
Here's what I see in the admin dashboard logs

```
...
'data.section': '644ac9d8e7d077320f6b250a', 'data.filter': nan, 'data.floor': nan, 'data.sensed_speed': nan, 'data.vaccuracy': nan, 'data.accuracy': nan, 'data.elapsedRealtimeNanos': nan, 'data.local_dt_year': nan, 'data.local_dt_month': nan, 'data.local_dt_day': nan, 'data.local_dt_hour': nan, 'data.local_dt_minute': nan, 'data.local_dt_second': nan, 'data.local_dt_weekday': nan, 'data.local_dt_timezone': nan, 'data.user_id': nan, 'data.metadata_write_ts': nan}], export_format='csv', page_current=0, page_size=50, filter_options={'case': 'sensitive'}, sort_action='native', sort_mode='single', style_table={'overflowX': 'auto'}, style_cell={'textAlign': 'left'})`
--
In general, Dash properties can only be
dash components, strings, dictionaries, numbers, None,
or lists of those.
```

shankari:
@achasmita do you see `nan` values in the trajectory table in your testing?

achasmita:
> @achasmita do you see `nan` values in the trajectory table in your testing?

Those all columns are empty on my table so it is working fine on my side.
![Screen Shot 2023-09-28 at 3 55 39 PM](https://github.com/e-mission/op-admin-dashboard/assets/79387860/d036fe1f-6340-4a04-8fe3-133233390948)

achasmita:
and I do not see nan values either

shankari:
maybe the nan values are just not displayed. Can you print out the last 5-6 rows to see if they do in fact have NaN values?

achasmita:
There is actually NaN values in previous dataset
![Screen Shot 2023-09-28 at 5 01 11 PM](https://github.com/e-mission/op-admin-dashboard/assets/79387860/51fa7fe8-bfd5-44a9-931d-c3f2533348c9)

shankari:
@achasmita

poking around on staging, this is what I see. If I try to load all trajectories, it takes a long time to load.
If I filter to only Sept (1/9 to 23/9), though, I get the same data as the previously loaded tab

### UUID table

[...]

PR #71, opened 2023-09-29
Fixed the issues with columns in trajectory table.

achasmita:


achasmita:
Testing is done and table is loading fine now.
Screen Shot 2023-09-29 at 12 24 06 AM
Screen Shot 2023-09-29 at 12 24 39 AM
Screen Shot 2023-09-29 at 12 25 19 AM
Screen Shot 2023-09-29 at 12 26 28 AM
Screen Shot 2023-09-29 at 12 26 49 AM

achasmita:
I am not able to make filter by date range yet ,  i will try it in morning again.

shankari:
I think that the code to filter by date range is fairly simple - you should be able to copy it from the trip code
https://github.com/e-mission/op-admin-dashboard/blob/35fc91dd245e65f086b2a9af4f6a4711cfaec659/utils/db_utils.py#L51

Let's try to get it in before merging because people may **want to** filter by a date range.

achasmita:
I am using the same code, and as table is loaded only when tab is selected, I am still figuring out on making both thing works.

achasmita:
Is it ok if we load a recent week data at first and then allow user to load data based on the date range they select. Instead of loading data only when trajectory table is selected ?

shankari:
@achasmita are you proposing to roll in the scalability fixes (https://github.com/e-mission/op-admin-dashboard/issues/51)?
I have no objections to that approach, but would like to understand why it is needed.

As I said before, **please record what you tried in the issue and how it failed**
"The goal is not just to get things to work, but to get them to work right".
https://github.com/e-mission/op-admin-dashboard/issues/70#issuecomment-1740268393

shankari:
@achasmita it will also give me a sense of how complicated the problem is, and whether we should go ahead with pushing this change

achasmita:
> @achasmita are you proposing to roll in the scalability fixes (#51)? I have no objections to that approach, but would like to understand why it is needed.
> 
> As I said before, **please record what you tried in the issue and how it failed** "The goal is not just to get things to work, but to get them to work right". [#70 (comment)](https://github.com/e-mission/op-admin-dashboard/issues/70#issuecomment-1740268393)

Yes it will work for both. I tried to pass the date range to filter trajectory table but the table was not loading as expected. It was loading all data everytime. I saw the date range change when i select different date range #72  but the data remains 
same.

achasmita:
> @achasmita are you proposing to roll in the scalability fixes (#51)? I have no objections to that approach, but would like to understand why it is needed.
> 
> As I said before, **please record what you tried in the issue and how it failed** "The goal is not just to get things to work, but to get them to work right". [#70 (comment)](https://github.com/e-mission/op-admin-dashboard/issues/70#issuecomment-1740268393)

Yes it will work for both. I tried to filter the trajectory table in previous code  by date range but there was no change in data even after filter is provided. Everytime all the data is loadaed in table so last time I filed the issue.

achasmita:
Ok I figured it out, I was doing sth wrong while passing the date, now it works both the way either we can load trajectory tab when respective tab is selected and apply filter range or we can load data for current week and use date range to load data as needed.

pages/data.py
```diff
@@ -4,7 +4,7 @@
 The workaround is to check if the input value is None.
 """
 from dash import dcc, html, Input, Output, callback, register_page, dash_table
-
+from datetime import date
```
shankari, line 7: I am merging this for now, but we should revisit the time handling in general since datetime does not handle timezones very well. We should use `arrow` like the rest of the e-mission codebase, and also consider date formatting etc.

Can you please file a follow-up issue for this?

shankari (approved):

PR #86, opened 2023-10-25
Update Docker image tag

MukuFlash03:
Fixed base image for server after addressing latest vulnerabilities in [this PR](https://github.com/e-mission/e-mission-server/pull/943).

Updated docker image tag so dashboard can pull from the latest base server image.


shankari:
For the record, PR https://github.com/e-mission/e-mission-server/pull/943 does not fix any vulnerabilities (they were already fixed by the blanket update) - it just fixes them more gracefully 😄

PR #89, opened 2023-11-06
changed demographic table to display multiple survey in subtabs

achasmita:


achasmita:
table with single survey:
Screen Shot 2023-11-05 at 2 55 52 PM

table with multiple surveys
Screen Shot 2023-11-05 at 7 53 49 PM
Screen Shot 2023-11-05 at 7 54 06 PM
Screen Shot 2023-11-05 at 7 54 16 PM
Screen Shot 2023-11-05 at 7 54 25 PM

pages/data.py
```diff
@@ -101,7 +113,31 @@ def render_content(tab, store_uuids, store_trips, store_demographics, store_traj
...
+        if data:
+            columns = list(data[0].keys())
+            has_perm = perm_utils.has_permission('data_demographics')
+
+        df = pd.DataFrame(data)
+        if df.empty or not has_perm:
+            return None
+
+        df = df.drop(columns=[col for col in df.columns if col not in columns])
+        df = clean_location_data(df)
```
shankari, line 137: why are we setting the demographic dataframe to a location dataframe?

utils/db_utils.py
```diff
@@ -104,23 +104,37 @@ def query_confirmed_trips(start_date, end_date):
 def query_demographics():
     logging.debug("Querying the demographics for (no date range)")
     ts = esta.TimeSeries.get_aggregate_time_series()
```
shankari, line 106: you should add a comment here highlighting the different return type here since it is not the same as the other dataframes.

shankari (changes requested):
Almost there, only need a few comments/cleanups

pages/data.py
```diff
@@ -77,8 +77,20 @@ def render_content(tab, store_uuids, store_trips, store_demographics, store_traj
...
     elif tab == 'tab-demographics-datatable':
         data = store_demographics["data"]
-        columns = list(data[0].keys())
-        has_perm = perm_utils.has_permission('data_demographics')
+        # if only one survey is available, process it without creating a subtab
+        if len(data) == 1:   
+            key =next(iter(data))
+            data = data[key]
+            columns = list(data[0].keys())
+            has_perm = perm_utils.has_permission('data_demographics')
```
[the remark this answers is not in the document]
achasmita: Yes, it there is only one survey it will instantly check and create table else it will check once subtab is created.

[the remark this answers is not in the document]
shankari: Right, but I don't see the need for this distinction
why can you not just check this outside the `if`?
Also the `has_perm` set in here is used in

```
 95     df = pd.DataFrame(data)
 96     if df.empty or not has_perm:
 97         return None
 98
```

so it is not clear how setting it in the callback for the demographic table subtabs helps in any way

Have you actually tried testing this? If you set the `data_demographics` permission to false, does it actually get hidden?

shankari (changes requested):

pages/data.py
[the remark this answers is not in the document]
shankari: > Yes, it doensnot matter, it functions same whether we keep it in or out of if.

I don't understand/believe this. the check for `has_perm` is in `render_content`.
Setting a variable in `populate_datatable` should not affect the check - `has_perm` is not a global variable.
If you claim that it does - can you explain why?

The way that you are testing is incorrect wrt end to end testing - you need actually use a config that has the field set.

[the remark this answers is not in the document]
achasmita: <details>

<summary>This function is  in permission.py:  </summary>

>Added this to check if has_perm is working and it gives same above result.
         ```permissions['data_demographics'] = [False]```

```  bash
STUDY_CONFIG = os.getenv('STUDY_CONFIG')
PATH = os.getenv('CONFIG_PATH')
CONFIG_URL = PATH + STUDY_CONFIG + ".nrel-op.json"
response = requests.get(CONFIG_URL)
config = json.loads(response.text)
surveyinfo = config.get("survey_info",
    {
      "surveys": {
        "UserProfileSurvey": {
          "formPath": "json/demo-survey-v2.json",
          "version": 1,
          "compatibleWith": 1,
          "dataKey": "manual/demographic_survey",
          "labelTemplate": {
            "en": "Answered",
            "es": "Contestada"
          }
        }
      },
      "trip-labels": "MULTILABEL"
    })
permissions = config.get("admin_dashboard", {})
permissions['data_demographics'] = [False]

def has_permission(perm):
    return False if permissions.get(perm) is False else True
```

</details>

pages/data.py
```diff
@@ -101,7 +113,30 @@ def render_content(tab, store_uuids, store_trips, store_demographics, store_traj
...
+)
+
+def update_sub_tab(tab, store_demographics):
+    data = store_demographics["data"]
+    if tab in data:
+        data = data[tab]
+        
+        if data:
+            columns = list(data[0].keys())
+            has_perm = perm_utils.has_permission('data_demographics')
```
shankari, line 130: @achasmita as we discussed, I don't see how setting this here affects anything in the main tab. you should remove this line. you might need to move the `has_perm` check in `render_context` before the `if` to compensate.

shankari (changes requested):

shankari (approved):

pages/data.py
```diff
@@ -77,8 +77,23 @@ def render_content(tab, store_uuids, store_trips, store_demographics, store_traj
         has_perm = perm_utils.has_permission('data_trips')
     elif tab == 'tab-demographics-datatable':
         data = store_demographics["data"]
-        columns = list(data[0].keys())
         has_perm = perm_utils.has_permission('data_demographics')
+        # if only one survey is available, process it without creating a subtab
+        if len(data) == 1: 
+            # here data is a dictionary 
+            data = next(iter(data.values()))
```
shankari, line 84: @achasmita I had commented on this in:
https://github.com/e-mission/op-admin-dashboard/pull/89#discussion_r1388628743

you removed it in 
https://github.com/e-mission/op-admin-dashboard/pull/89/commits/53dafd6af1a2f4d36f0cabd49b3418fc24973539

but now that I check the code, this is still here.

Please fix.

utils/db_utils.py
```diff
@@ -102,25 +102,40 @@ def query_confirmed_trips(start_date, end_date):
...
-            if col in df.columns:
-                df.drop(columns= [col], inplace=True) 
-    return df
 
+    entries = ts.find_entries(["manual/demographic_survey"])
+    data = list(entries)
+
+    available_key = {}
+    for entry in data:
+        survey_key = next(iter(entry['data']['jsonDocResponse']))
```
shankari, line 114: Ditto

shankari (changes requested):

shankari:
@achasmita please fill in "testing done"

shankari:
Going to merge this now.
Note that I am squash-merging this, so please handle it properly while merging changes from upstream.

PR #90, opened 2023-11-29
load only  last 7 days data for trip and trajectory table at begining.

achasmita:


achasmita:
- > Trip table will load data for last 7 days initially 

- > Trajectory table load data for last 7 days by default when trajectory tab is selected for now
 >> Both tables are initially empty as we dont have any data currently for past 7 days
Screen Shot 2023-11-29 at 1 34 49 PM
Screen Shot 2023-11-29 at 1 35 01 PM

>> Changing it to load for past 41 days
Screen Shot 2023-11-29 at 8 32 42 PM
Screen Shot 2023-11-29 at 8 32 54 PM

achasmita:
> > As we are reading data only for last 7 days on login, I changed the trips trend on initial screen(home screen) to "last week trip trend".

<details>

<summary>

I noticed that when we change date range in date filter,  trip trends also get changed based on date range in filter.  Previously I renamed it to "last week trip trend" as it will show last week trip trend in begining.  But as it gets updated with change in date filter we need to think some other name. 

</summary>

Screen Shot 2023-12-01 at 1 01 42 PM


</details>

shankari:
@achasmita this PR is still a draft, but it is in my "to review" bucket. What input (if any) would you like from me here?

achasmita:
<details>

<summary> I was thinking what name can we give for trip trend as it will display trend for a week at beginning and can be changed to display trend for days, months or years. What if we include the date range itself in trip trend like this: </summary>



Screen Shot 2023-12-02 at 11 39 48 AM

</details>

shankari:
>  I was thinking what name can we give for trip trend as it will display trend for a week at beginning and can be changed to display trend for days, months or years. What if we include the date range itself in trip trend like this: 

I think this is fine for now. We can always tweak it later based on feedback.

shankari (approved):
LGTM!

PR #91, opened 2023-12-11
added mode specific filter for maps

achasmita:


achasmita:
> Adding mode specific filter to maps

<details>

<summary> I started grouping data by confirmed modes and used that data to filter all three maps: </summary>

Screen Shot 2023-12-11 at 11 27 53 AM

Screen Shot 2023-12-11 at 11 28 23 AM

Screen Shot 2023-12-11 at 11 29 22 AM


</details>

achasmita:
Here I have grouped data by mode, so that data is plotted based on selected modes:
> Bubble map:
>> Before: 
Screen Shot 2023-12-13 at 7 04 11 PM

>>After mode filter is added: 
Screen Shot 2023-12-13 at 7 09 30 PM
Screen Shot 2023-12-13 at 7 09 41 PM


>Density Heatmap:
>>Before:
Screen Shot 2023-12-13 at 7 04 24 PM

>>After mode filter is added:
Screen Shot 2023-12-13 at 7 10 03 PM

>Trip Lines map:

>> Before: 
Screen Shot 2023-12-13 at 7 05 39 PM

>>After mode filter is added
Screen Shot 2023-12-13 at 7 21 58 PM
Screen Shot 2023-12-13 at 7 22 24 PM

achasmita:
With the new changes
> We can now filter bubble and density heat map by different modes (there were no filters in this two maps before).
> And for trip map as there was already a filter option with UUID's and emails:
>>now it can be either filtered by mode or by UUID's and emails
>>
Screen Shot 2023-12-17 at 9 41 20 PM
>>
Screen Shot 2023-12-17 at 9 41 50 PM

pages/map.py
```diff
@@ -148,6 +196,15 @@ def create_user_emails_options(trips_group_by_user_id):
             options.append(create_single_option(user_email, color))
     return options, user_emails
 
+def create_user_modes_options(trips_group_by_user_mode):
+    options = list()
+    user_modes = set()
+    for user_mode in trips_group_by_user_mode:
+        color = trips_group_by_user_mode[user_mode]['color']
```
shankari, line 203: I am looking at the implementation of `get_trips_by_user_mode`, and I don't see where the 'color' is filled in.
`trips_group_by_user_mode = trips_df.groupby('data.user_input.mode_confirm')` so unless `trips_df` has a `color`, `trips_group_by_user_mode` will not have a color.

Can you please clarify how this is supposed to work?

pages/map.py
```diff
@@ -198,7 +259,7 @@ def create_user_emails_options(trips_group_by_user_id):
     Input('user-id-dropdown', 'value'),
 )
 def update_user_ids_options(trips_data, selected_user_ids):
-    user_ids_options, user_ids = create_user_ids_options(trips_data['users_data'])
+    user_ids_options, user_ids = create_user_ids_options(trips_data[0]['users_data'])
```
shankari, line 262: why do we have a hardcoded `trips_data[0]` here?
Are we only plotting for the first user?

pages/map.py
```diff
@@ -268,9 +348,17 @@ def store_trips_map_data(trips_data):
...
             trips = trips_group_by_user_id.get_group(user_id).sort_values('trip_start_time_str').to_dict("records")
             users_data[user_id] = {'color': color, 'trips': trips}
-            for trip in trips:
-                coordinates['lon'].append(trip['start_coordinates'][0])
-                coordinates['lon'].append(trip['end_coordinates'][0])
-                coordinates['lat'].append(trip['start_coordinates'][1])
-                coordinates['lat'].append(trip['end_coordinates'][1])
-    return {'users_data': users_data, 'coordinates': coordinates}
+    groupped_data = []
+    groupped_data.append({'users_data':users_data})
```
shankari, line 352: nit: `grouped`

pages/map.py
```diff
@@ -268,9 +348,17 @@ def store_trips_map_data(trips_data):
...
+    if trips_group_by_user_mode:
+        user_modes = list(trips_group_by_user_mode)
+        n = len(user_modes) % 360
+        k = 359 // (n - 1) if n > 1 else 0
+        for ind, user_mode in enumerate(trips_group_by_user_mode.groups.keys()):
+            color = f'hsl({ind * k}, 100%, 50%)'
+            trips = trips_group_by_user_mode.get_group(user_mode).sort_values('trip_start_time_str').to_dict("records")
+            users_data[user_mode] = {'color': color, 'trips': trips}
+    groupped_data.append({'users_data':users_data})
+    return groupped_data
```
shankari, line 364: Isn't this almost the same as lines 343 to 350. Why is this not DRYed up?

pages/map.py
shankari, line 352: why are you combining the users_data from the user_id groups with the one from the modes?
Is `grouped_data` going to have two separate entries called `users_data`?
Do you need both? How do you know which is which?
If you do need both, why not make it clearer how to distinguish between them?

pages/map.py
```diff
@@ -211,31 +272,48 @@ def update_user_ids_options(trips_data, selected_user_ids):
...
+    user_modes=set(selected_user_modes) if selected_user_modes is not None else set()
     if selected_user_emails is not None:
         for user_email in selected_user_emails:
             user_ids.add(str(ecwu.User.fromEmail(user_email).uuid))
-
+    arg1 = trips_data[0].get('users_data', {})
+    arg2 = user_ids       
+    if(selected_user_modes):
+        arg1 = trips_data[1].get('users_data', {})
+        arg2 = user_modes
```
shankari, line 310: why are these variables called `arg1` and `arg2`?
If they do need to be called those names, why are they not used consistently?

pages/map.py
```diff
@@ -57,45 +57,85 @@ def create_lines_map(trips_group_by_user_id, user_id_list):
...
+    for user_mode in user_mode_list:
+        color = trips_group_by_user_mode[user_mode]['color']
+        trips = trips_group_by_user_mode[user_mode]['trips']
+
+        for trip in trips:
+            coordinates['lon'].append(trip['start_coordinates'][0])
+            coordinates['lon'].append(trip['end_coordinates'][0])
+            coordinates['lat'].append(trip['start_coordinates'][1])
+            coordinates['lat'].append(trip['end_coordinates'][1])
+            coordinates['color'].extend([color,color])
```
shankari, line 119: why is this copied over twice?

shankari (changes requested):
@achasmita please provide some additional context for your changes. We don't just want code, we want maintainable code. There are multiple copy-pasted lines, and weird variable names. I checked to see if the commits had more information, but each commit only has a single line "added mode specific filter for maps "

I am not sure what that means.
Please clean up and document the code and resubmit for review.

achasmita:
<Details>

<summary>
All three maps can now be filtered using modes:
</summary>

> ### Map Type

1. **Trip Lines**

  > - Trip lines used to be previously filtered using 'User UUIDs' and 'User Emails'.
  >> 

  > - Now a new filter 'Modes' is added
  > - Now we can filter the map with either '**User UUIDs** and **User Emails**' or '**Modes**'.
  >> Screen Shot 2024-01-10 at 12 42 41 PM
  >> Screen Shot 2024-01-10 at 12 43 20 PM




2. **Bubble Map** and **Density Heatmap**

> - There were no filter for Bubble Map and Density Heatmap before.
>>
>>

> - Now a filter 'Modes' is added:
            For now, it functions similarly to the Trip lines map, where coordinates will be plotted once the user selects the modes.
            Alternatively, We can continue to plot them as before: once the map is selected, all co-ordinates will be plotted with different colors corresponding to the modes.
>> Screen Shot 2024-01-10 at 12 52 25 PM
>> Screen Shot 2024-01-10 at 12 52 42 PM

To filter data by modes:

- Started with grouping data by modes using a function ```get_trips_group_by_user_mode```
```
def get_trips_group_by_user_mode(trips_data):
    trips_group_by_user_mode = None
    trips_df = pd.DataFrame(trips_data['data'])
    if not trips_df.empty:
        trips_df['data.user_input.mode_confirm'] = trips_df['data.user_input.mode_confirm'].fillna('Unknown')
        trips_group_by_user_mode = trips_df.groupby('data.user_input.mode_confirm')
    return trips_group_by_user_mode
```
- Edited function ```store_trips_map_data```:

> - created new function ``` process_trips_groups``` so it can be used for both  ```trips_group_by_user_id``` and ```trips_group_by_user_mode```.
```
users_data = dict()
    if trips_group:
        keys = list(trips_group)
        n = len(keys) % 360
        k = 359 // (n - 1) if n > 1 else 0
        for ind, key in enumerate(trips_group.groups.keys()):
            color = f'hsl({ind * k}, 100%, 50%)'
            trips = trips_group.get_group(key).sort_values('trip_start_time_str').to_dict("records")
            users_data[key] = {'color': color, 'trips': trips}  
    return users_data

```
>> This code block  was previously inside ```store_trips_map_data``` (it processes a group of trips, assigns color to each group, sorts the trips within each group and stores the processed data in a dictionary). 
</Details>

shankari:
@achasmita 

Please see my comment above:
>  I checked to see if the commits had more information, but each commit only has a single line "added mode specific filter for maps "

I still only see one line for this comment "cleaned and made changes in previous code"
I expect to see an actual description of the changes, similar to https://github.com/e-mission/em-public-dashboard/pull/112/commits/68911815a6b3168f9c48fcd0169f2e64fb0d17be

Please confirm that you have seen this message and will implement it going forward

achasmita:
> @achasmita
> 
> Please see my comment above:
> 
> > I checked to see if the commits had more information, but each commit only has a single line "added mode specific filter for maps "
> 
> I still only see one line for this comment "cleaned and made changes in previous code" I expect to see an actual description of the changes, similar to [e-mission/em-public-dashboard@6891181](https://github.com/e-mission/em-public-dashboard/commit/68911815a6b3168f9c48fcd0169f2e64fb0d17be)
> 
> Please confirm that you have seen this message and will implement it going forward

I will make sure to follow it going forward

pages/map.py
```diff
@@ -57,45 +57,78 @@ def create_lines_map(trips_group_by_user_id, user_id_list):
...
```
shankari, line 71: Can you clarify why you picked this data structure, which requires you to iterate over all the trips? Also it seems weird that you concatenate the start and end longitudes and latitudes separately. Are you sure this is the expected input? Where is the related documentation?

shankari: Note to self: this is copied over from
https://github.com/e-mission/op-admin-dashboard/pull/91/files#diff-eed798ce43b7a089fbd9491f30276d8779a357c64583d343e17b025d0c1559eeL271

I would still like to see some validation of this now that we are not in a hurry to get the admin dashboard merged.

pages/map.py
```diff
@@ -57,45 +57,78 @@ def create_lines_map(trips_group_by_user_id, user_id_list):
...
+    return coordinates
+
+
+def create_heatmap_fig(trips_group_by_user_mode, user_mode_list):
+    coordinates = get_map_coordinates(trips_group_by_user_mode, user_mode_list)
```
shankari, line 76: Given that you have a consistent pattern across call the `create_XXX_fig`, with the first line being

```
coordinates = get_map_coordinates(trips_group_by_user_mode, user_mode_list)
```

it seems like you can just pass in the coordinates directly. Then you don't have to iterate over all the trips every time the user switches between map types.

pages/map.py
```diff
@@ -57,45 +57,78 @@ def create_lines_map(trips_group_by_user_id, user_id_list):
...
-                lat=data['lat'],
+                lon=coordinates['lon'],
+                lat=coordinates['lat'],
+                name = '',
+                
             )
         )
+
+        fig.add_trace(
+            go.Scattermapbox(
```
shankari, line 89: why are we displaying a `Scattermapbox` in `create_heatmap_fig`?
We already have one in `create_bubble_fig` below.

shankari: And why is this not showing up in the "Testing done"? I checked the heatmaps in the most recent set of tests, and I don't see any bubbles in there. This makes me skeptical of the testing done.

pages/map.py
```diff
@@ -244,33 +312,39 @@ def update_output(map_type, selected_user_ids, selected_user_emails, trips_data)
...
     return disabled, disabled
 
 
+def process_trips_group(trips_group):
+    users_data = dict()
+    if trips_group:
+        keys = list(trips_group)
+        n = len(keys) % 360
+        k = 359 // (n - 1) if n > 1 else 0
+        for ind, key in enumerate(trips_group.groups.keys()):
```
shankari, line 332: Can you please add the explanation

> This code block was previously inside store_trips_map_data (it processes a group of trips, assigns color to each group, sorts the trips within each group and stores the processed data in a dictionary).

to the code?

Not everybody is going to look at the related PR, especially when the information is hidden in a detail

pages/map.py
```diff
@@ -244,33 +312,39 @@ def update_output(map_type, selected_user_ids, selected_user_emails, trips_data)
...
+            color = f'hsl({ind * k}, 100%, 50%)'
+            trips = trips_group.get_group(key).sort_values('trip_start_time_str').to_dict("records")
```
shankari, line 334: why do we need to sort the trips? Given a map-based visualization, there is effectively no ordering among trips.

shankari (changes requested):

pages/map.py
achasmita: It is sorting the trip based on their start time, but as we are plotting all the co-ordinates collectively there is no advantage of sorting here. I will skip the sorting part.

pages/map.py
achasmita: I added  it just to differentiate when different modes are selected. I have attached a screenshot of heatmap with bubble above.

shankari: I am not sure this is useful, and I am not even sure that users will notice the different colored bubbles in the heatmap. I think that just displaying the heatmap that includes all the trips selected is sufficient. @JGreenlee thoughts?

shankari:
Pending feedback from @JGreenlee and potential removal of the scatterplot in the heatmap.

pages/map.py
JGreenlee: Since the heatmap already uses color as an attribute for intensity, we cannot easily convey any other information using color.
I would suggest removing the bubbles from the heatmap.

> I think that just displaying the heatmap that includes all the trips selected is sufficient.

I agree. This is sufficient as long as modes can still be added/removed and the heatmap reflects those changes.

JGreenlee: @achasmita

shankari:
@achasmita please add testing done after the last set of changes

achasmita:
Scatter plot is removed from heatmap and testing is done.
Screen Shot 2024-03-05 at 9 29 29 PM

pages/map.py
shankari: @achasmita I don't think this has been addressed yet. Why do you pick this data structure which seems like it has a lot of duplicated effort to copy things around?

shankari (approved):
I see one cleanup that has not been addressed, but in the interests of time, I'm merging this and moving it to "cleanup required" @jiji14 you can address this in your task list.

Issue #73, opened 2023-10-02
Need to fix time handling for filter by date-range


achasmita:
Since datetime does not handle timezones very well. We should use arrow like the rest of the e-mission codebase, and also consider date formatting etc.

shankari:
@jgreenlee maybe you can tackle this

JGreenlee:
The UX fixes for the datepicker itself were quick and I implemented those on https://github.com/e-mission/op-admin-dashboard/pull/96

Swapping out `datetime` for `arrow` seems slightly more involved (I can see it was attempted before) and I'd like a bit more information.

Is there anything that is currently broken with the current implementation using `datetime`? Or is it just because we want to be consistent with e-mission-server?

shankari:
@JGreenlee I think it is a bit of both. using arrow is consistent with e-mission-server, and, as indicated above, it also handles timezones better. If I select Jan 1st to Jan 10th, should I expect to see a trip back home after an NYE party? I would think so, but to ensure that it happens, I think we need to handle timezones correctly

JGreenlee:
Suppose I'm a program administrator in California and a user in the UK has a trip on Jan 1 from 1am-2am.
I select the date range of Jan 1 - Jan 10 and that user's trip shows up. When that trip occurred, it was still Dec 31st in California time.

In other words, the timezone of the program administrator does not matter; all that matters is what time it was for the user when their trip occurred.

Is that correct?

JGreenlee:
The problem is that we're currently matching by timestamps (which are Unix time and always relative to 0 seconds GMT).

To accomplish the above NYE scenario, we'd need to support a different kind of matching on `e-mission-server` (ie matching by calendar date in local time).

shankari:
Right, I created the `local_date` fields to help with this

JGreenlee:
Can we just reformat to YYYYMMDD while doing the gt / lt comparison ?

JGreenlee:
Either way, as a first step I'm currently working on adding `arrow` as a drop-in replacement for `datetime`.

Then a second step can handle timezones once we decide how to proceed (maybe rework UI, maybe server changes)

shankari:
> Can we just reformat to YYYYMMDD while doing the gt / lt comparison ?

I think so (would have to think through and test to make sure that it works for all cases). *BUT* the database does not actually store data in YYYYMMDD - the fmt_time does store the data as a string, but does so as YYYYMMDDThh:mm:ss 
[...]

PR #96, opened 2024-01-18
📅 Datepicker: set initial range, use unambiguous date format, use `arrow`, dropdown for `timezone` option

JGreenlee:

### DIFF
--- a/app_sidebar_collapsible.py
+++ b/app_sidebar_collapsible.py
@@ -23,6 +23,7 @@ import logging
 if os.getenv('DASH_DEBUG_MODE', 'True').lower() == 'true':
     logging.basicConfig(level=logging.DEBUG)
 
+from utils.datetime_utils import iso_to_date_only
 from utils.db_utils import query_uuids, query_confirmed_trips, query_demographics
 from utils.permissions import has_permission
 import flask_talisman as flt
@@ -204,9 +205,7 @@ app.layout = make_layout
     Input('date-picker-timezone', 'value'),
 )
 def update_store_uuids(start_date, end_date, timezone):
-    # trim the time part, leaving only date as YYYY-MM-DD
-    start_date = start_date[:10] if start_date else None
-    end_date = end_date[:10] if end_date else None
+    (start_date, end_date) = iso_to_date_only(start_date, end_date)
     dff = query_uuids(start_date, end_date, timezone)
     records = dff.to_dict("records")
     store = {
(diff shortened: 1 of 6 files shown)


JGreenlee:
Here is what the datepicker now shows on the initial load:

image

JGreenlee:
Switched to using `arrow` instead of `datetime`.

We do not `import` from `datetime` anymore, although the `datetime` object is still used internally (by `pandas` and by `arrow` itself)

JGreenlee:
Added a dropdown for UTC time vs. Local timezone (as described in https://github.com/e-mission/op-admin-dashboard/issues/73#issuecomment-1908770258)

UI now looks like this:

<table>
<td>
image
</td><td>
image
</td></table>

JGreenlee:
As it turns out, the datepicker does not keep current to "today's" date for the selection.
Today, a user reported that the datepicker only allowed them to retrieve data up to January 15 (despite the date today being January 31).

The reason for this is to do with how Plotly Dash works – the DatePicker component gets generated once when the dashboard is deployed, and does not update on every page load. Our last production deployment was Jan 15, so that is what it is locked to.

Described in https://community.plotly.com/t/refreshing-the-start-date-everyday-for-datepickersingle/22234

According to https://dash.plotly.com/live-updates ("Updates on Page Load"), if we define the page content as a function rather than a variable, it will be generated on each page load / refresh.

JGreenlee:
Fixed in d9c7df2915431b9fd0258e4502e0bd31a2083a3e

To prove that it was actually updating and for the purpose of testing, I injected a simple `Span` with a fully printed out datetime.

Every time I refresh, the 'seconds' updates, which proves that the controls are being re-computed.
Screen Recording 2024-02-02 at 1 59 11 AM (1)

app_sidebar_collapsible.py
```diff
@@ -207,15 +238,13 @@ def update_store_uuids(start_date, end_date):
...
-        end_date_obj = date.today()
-        start_date_obj = end_date_obj - timedelta(days=7)
-    else:
-        start_date_obj = date.fromisoformat(start_date) 
-        end_date_obj = date.fromisoformat(end_date)
-    df = query_confirmed_trips(start_date_obj, end_date_obj)
+def update_store_trips(start_date, end_date, timezone):
+    # trim the time part, leaving only date as YYYY-MM-DD
+    start_date = start_date[:10] if start_date else None
+    end_date = end_date[:10] if end_date else None
```
shankari, line 246: this code is repeated multiple times and should be pulled out into a separate function.
This is not just for lowering the lines of code (from 2 to 1) but consolidating the calls so that if we ever change the display/parse method, we can do so in one place.

shankari (changes requested):
LGTM other that the single refactor below.

app_sidebar_collapsible.py
JGreenlee: That would be much better! I didn't see any existing files where a function like that would make sense to go in.
So to refactor, I think I will create a new file `utils/datetime_utils.py` with this new function, and maybe also the `get_ts_range` function (which I had created in `db_utils`)

JGreenlee: refactored

shankari (approved):

PR #97, opened 2024-01-20
Support both humanized and raw values in trips table.

achasmita:


achasmita:
Now trips table includes dropdowns which help user to display, hide and export raw value.
Screen Shot 2024-01-20 at 10 03 18 AM
Screen Shot 2024-01-20 at 10 03 35 AM
Screen Shot 2024-01-20 at 10 03 57 AM

shankari:
@achasmita can you please ensure that this PR only includes the humanized/non humanized value changes?
Also, I asked for a link to the documentation on hidden columns, and for feedback on visual design from @JGreenlee

utils/constants.py
```diff
@@ -17,12 +17,29 @@
...
     "data.duration",
+    "data.duration.raw",
     "data.distance",
+    "data.distance.raw",
     "data.start_loc.coordinates",
     "data.end_loc.coordinates",
     "user_id"
 ]
 
+INITIAL_TRIP_COLS =[
```
shankari, line 28: why are we essentially duplicating `VALID_TRIP_COLS` here? I don't love the listing out of valid cols (so many hardcoded lists), but there is really no reason to have two almost identical lists.

Also, for raw, we should just provide them with everything that we have in the table. No filtering.

utils/db_utils.py
```diff
@@ -85,12 +85,14 @@ def query_confirmed_trips(start_date, end_date):
         # https://github.com/e-mission/op-admin-dashboard/issues/29#issuecomment-1530105040
         # https://github.com/e-mission/op-admin-dashboard/issues/29#issuecomment-1530439811
         # so just replacing the distance and duration with the humanized values for now
+        df['data.distance.raw'] = df['data.distance']
```
shankari, line 88: IMHO, we should have `data.distance` as the raw value and `data.distance_miles` as humanized.

utils/db_utils.py
```diff
@@ -85,12 +85,14 @@ def query_confirmed_trips(start_date, end_date):
...
         use_imperial = perm_utils.config.get("display_config",
             {"use_imperial": False}).get("use_imperial", False)
         # convert to km to humanize
         df['data.distance'] = df['data.distance'] / 1000
         # convert km further to miles because this is the US, Liberia or Myanmar
         # https://en.wikipedia.org/wiki/Mile
+        df['data.duration.raw'] = df['data.duration']
```
shankari, line 95: Ditto

pages/data.py
```diff
@@ -140,13 +161,28 @@ def update_sub_tab(tab, store_demographics):
 
         df = df.drop(columns=[col for col in df.columns if col not in columns])
 
-        return populate_datatable(df)
-      
-def populate_datatable(df):
+        return populate_datatable(df, 'my-table')
```
shankari, line 164: Can we please use a more meaningful name than `my-table`?

pages/data.py
```diff
@@ -140,13 +161,28 @@ def update_sub_tab(tab, store_demographics):
...
+
+@callback(
+    Output('my-trips-table', 'hidden_columns'),
```
shankari, line 167: and `my-trips-table`

shankari (changes requested):
We should also provide humanized/raw values for the trajectory table.

utils/db_utils.py
achasmita: What if we have data.distance_meters as the raw value and data.distace as humanized. The humanized can be km or miles based on config.

shankari: What is the rationale for this suggestion?

JGreenlee:
I pulled this branch and tested it to evaluate UX.

<table><tr><td>
image
</td><td>
image
</td></tr></table>

I think it is not very clear to the user what the dropdown menu does. When the user makes a selection, they expect to get some feedback; it seems like the displayed table should update, but it does not.
I am assuming the dropdown selection only applies to the exported CSV files, but I don't think users will immediately understand that.

If a dropdown is used here, I think it's important for the displayed table to immediately update reflecting this change.

Otherwise, if the options will only apply to the exported file, I think a different mechanism for specifying these options would make more sense. Perhaps a small panel with:
- the title "Export Options"
- checkbox / checkboxes for "Raw values"
- the "Export" button.

JGreenlee:
I am also wondering: is it necessary for there to be separate options for each column?

Is there any reason someone would want raw distance with humanized duration? Or humanized distance with raw duration?

If not, this can be as simple as one toggle. Either *both* 'raw' or *both* 'humanized'

achasmita:
>If a dropdown is used here, I think it's important for the displayed table to immediately update reflecting this change.

The table should immediately update when user select the dropdown option. The new column is added to the end of the table.

shankari:
> If not, this can be as simple as one toggle. Either both 'raw' or both 'humanized'

I like this approach as well. Sometimes too much choice can be confusing ("paradox of choice")
We could start with that and make it more complex later

JGreenlee:
> The table should immediately update when user select the dropdown option. The new column is added to the end of the table.

Ah, I see now. I had been expecting it would replace the original column.

<hr>

If a simple toggle is sufficient, I think a [ToggleSwitch](https://dash.plotly.com/dash-daq/toggleswitch) would work best.
And if the intention is to append new columns at the end instead of replacing the originals, the text for this switch could be something like "Include columns for raw values".
And if possible, I think inserting the 'raw' columns immediately after their counterparts would be better rather than putting them all the way at the end.

I also think it would help very much if the columns had their units labeled. Currently, `data.distance` gives no indication of whether it's miles or kilometers. And `data.distance.raw` doesn't indicate that it's meters.

achasmita:
I liked the idea of switching between either both humanized column and both raw column instead of selecting each column from dropdown.
We can use a single button to switch instead of dropdowns:
>
Screen Shot 2024-01-22 at 9 16 11 PM

And the columns get switch once the button are clicked and so does the text on button, This way we can display table either with humanized value or with raw value:
>I still need to change the columns name
Screen Shot 2024-01-22 at 9 16 45 PM
Screen Shot 2024-01-22 at 9 17 23 PM

achasmita:
Screen Shot 2024-01-23 at 1 32 39 PM
Screen Shot 2024-01-23 at 1 33 06 PM

shankari (changes requested):
@achasmita I think that the change itself is likely fine, but there are a lot of additional changes which have no explanations. I would like to start enforcing clear documentation and code explanations.

Please add them (even if just as comments in the PR) and resubmit for review.

pages/data.py
```diff
@@ -3,13 +3,14 @@
 Since the dcc.Location component is not in the layout when navigating to this page, it triggers the callback.
 The workaround is to check if the input value is None.
 """
-from dash import dcc, html, Input, Output, callback, register_page, dash_table
+from dash import dcc, html, Input, Output, callback, register_page, dash_table, State
```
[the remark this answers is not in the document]
achasmita: I used state to capture the button_label, button label changes from ```Display column with humanized units``` to ```Display column with raw units``` based on number of clicks.

pages/data.py
```diff
@@ -75,6 +76,24 @@ def render_content(tab, store_uuids, store_trips, store_demographics, store_traj
             col['label'] for col in perm_utils.get_allowed_named_trip_columns()
         )
         has_perm = perm_utils.has_permission('data_trips')
+        df = pd.DataFrame(data)
+        if df.empty or not has_perm:
+            return None
+
+        df = df.drop(columns=[col for col in df.columns if col not in columns])
+        df = clean_location_data(df)
```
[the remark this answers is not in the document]
achasmita: Trips table is returned as part of the 'html.Div' but common filtering is performed afterward.

pages/data.py
```diff
@@ -116,9 +135,8 @@ def render_content(tab, store_uuids, store_trips, store_demographics, store_traj
         return None
 
     df = df.drop(columns=[col for col in df.columns if col not in columns])
-    df = clean_location_data(df)
 
-    return populate_datatable(df)
+    return populate_datatable(df, 'datatable')
```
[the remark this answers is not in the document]
achasmita: The string is just the table_id to pass some hidden columns inside datatable, I have used trips-datatable for trips table, other table share same filter so I used datatable for other tables.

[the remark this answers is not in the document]
shankari: I don't understand this comment. where is this used to pass "hidden columns inside datatable"? I have searched through the code, and I don't see any other instances of this being used. If this is not used, does it need to be passed in?

Also, it is not clear to me if you used the same table_id for multiple tables. But if you did, why? Why not use separate names for each datatable to clarify the code instead of leading to potential issues with reusing the table ID.

pages/data.py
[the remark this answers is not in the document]
shankari: I don't understand what this means. which common filtering? why does it need to be common? why does it need to be "afterward"?

shankari (changes requested):

pages/data.py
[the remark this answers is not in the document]
achasmita: I will go ahead and update the name, so there is no more confusion.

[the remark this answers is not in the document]
shankari: @achasmita updating the name only addresses the second part of this comment. I still don't see an answer to the first part.

pages/data.py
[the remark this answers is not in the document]
shankari: @achasmita this has not yet been answered

pages/data.py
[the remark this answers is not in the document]
achasmita: I have passed hidden columns on ```trips-datatable``` only. To pass the hidden column to trips table we need to provide id of the table and that is why I updated ```trips-datatable``` for trips table and ```datatable``` for other remaining table. And now have changed id for all table to make it more clear.

shankari:
@achasmita
- this is not yet answered: https://github.com/e-mission/op-admin-dashboard/pull/97#discussion_r1477604284
- wrt https://github.com/e-mission/op-admin-dashboard/pull/97#discussion_r1477659317
   - thank you for the clarification "To pass the hidden column to trips table we need to provide id of the table" But it still doesn't answer my question. Why do you need the trip_id for tables that don't have hidden columns? Can't those invocations remain changed?

achasmita:
>    Why do you need the trip_id for tables that don't have hidden columns? Can't those invocations remain changed?

Yes it will work if we pass a common id inside dash_table itself, and use same id to reference to trips table . That way we donot need to assign id for other tables.
```
       def populate_datatable(df):
             if not isinstance(df, pd.DataFrame):
                  raise PreventUpdate
             return dash_table.DataTable(
                 id= 'dash-table',
                 # columns=[{"name": i, "id": i} for i in df.columns],
                 data=df.to_dict('records'),

```
I think it looks better than making change in all tables.

pages/data.py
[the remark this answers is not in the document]
shankari: Ok that is more clear now. But that also implies that the clear_location_data should have been here all along - calling it as part of the common set of operations was invoking it unnecessarily many times.

pages/data.py
```diff
@@ -141,12 +160,36 @@ def update_sub_tab(tab, store_demographics):
...
+    columns = perm_utils.get_allowed_trip_columns()
+    columns.update(
+        col['label'] for col in perm_utils.get_allowed_named_trip_columns()
+    )
+    if n_clicks % 2 == 0:
+        hidden_col = ["data.duration_seconds", "data.distance_meters","data.distance"]
+        button_label = 'Display columns with raw units'
+    else:
+        hidden_columns = ["data.duration", "data.distance_miles", "data.distance_km", "data.distance"]
+        hidden_col = [col for col in hidden_columns if col in columns]
```
shankari, line 182: why do we need both `hidden_columns` and `hidden_col` in the `else` but not in the `if`
it seems like we should check, in both cases, whether the hidden columns exist

pages/data.py
```diff
@@ -141,12 +160,36 @@ def update_sub_tab(tab, store_demographics):
...
+    #return the list of hidden columns and the updated button label
+    return hidden_col, button_label
+
+
 def populate_datatable(df):
     if not isinstance(df, pd.DataFrame):
         raise PreventUpdate
     return dash_table.DataTable(
-        # id='my-table',
+        id= 'data-table', #common id for all table(users,trips,demographics,trajectories)
```
shankari, line 192: ok, so now I am completely confused. Per this comment: https://github.com/e-mission/op-admin-dashboard/pull/97#discussion_r1477659317 you need to have a table-id for the trips table to hide columns. So how does it work to hardcode the same table_id for all tables?

If you read my comments: https://github.com/e-mission/op-admin-dashboard/pull/97#discussion_r1477182533
I still don't see where you indicate to `update_dropdowns_trips` which table ID it is used for. And my question was 

Finally, my comment:
https://github.com/e-mission/op-admin-dashboard/pull/97#issuecomment-1926290830

> Why do you need the trip_id for tables that don't have hidden columns? Can't those invocations remain changed?

is unchanged.

You used to have all the `trip_id`s be `datatable`. They are now `data-table`. Why is this better? And to return to my fundamental question, do you need to have a `table_id` for the tables that don't have hiddent columns?

shankari (changes requested):

pages/data.py
achasmita: It will not make any difference, most of the id's  provided for tab and buttons are in that format so I just changed it. 

We call same function ```populate_datatable``` to create all the datatables, so either we should provide id for all the table or none of them.

shankari: @achasmita You have not answered the first part of my question

> ok, so now I am completely confused. Per this comment: https://github.com/e-mission/op-admin-dashboard/pull/97#discussion_r1477659317 you need to have a table-id for the trips table to hide columns. So how does it work to hardcode the same table_id for all tables?

Please read the question carefully and answer comprehensively

achasmita: > Sorry for confusion, there were some extra line of code from previous implementation while using dropdown previously. I have cleaned it now.

<details>

<summary>  

Here I have summarized the changes in trips table:  

</summary>

The ```update_dropdowns_trips``` callback function is responsible for updating the hidden column based on user interaction with the ```Display column with raw units``` button
  
```bash
  @callback(
      Output('data-table', 'hidden_columns'), 
      Output('button-clicked', 'children'), 
      Input('button-clicked', 'n_clicks'), 
      State('button-clicked', 'children') 
  )
  def update_dropdowns_trips(n_clicks, button_label):
      if n_clicks % 2 == 0:
          hidden_col = ["data.duration_seconds", "data.distance_meters","data.distance"]
          button_label = 'Display columns with raw units'
      else:
          hidden_columns = ["data.duration", "data.distance_miles", "data.distance_km", "data.distance"]
          button_label = 'Display columns with humanzied units'
      return hidden_col, button_label
  ```
  It consists of two outputs:
  - hidden_colums
    - Updates the list of hidden columns in the trips table.
    - Controls which columns are visible to the user
  I got the idea of hidden_columns from: https://dash.plotly.com/datatable/interactivity
  
  - button-clicked
    - Based on number of clicks it update the label of the butoon to either ```Display column with raw units``` or  ```Display column with humanized units```
  
  > We need id for table in this callback to specify where the change need to be updated(hidden_columns).
  > We have single function to poputate datatable ```def populate_datatable(df)``` which we is used to populate all datatable and it has single parameter ```df```.
  -  To pass ```id``` to table either I have to use  arguments of dash_table.DataTable ```id``` 
      - where i can directly provide id or pass it as a  new parameter from  ```def populate_datatable(df, table_id)```
  
  Here, I am using common id because the  ```update_dropdowns_trips``` is triggered only when we are inside trip table and interact with button so it will not affect any other table. If it still looks confusing, we can assign unique id to each table.
  
</details>

shankari: @achasmita I am not asking you to change the code. My question is this:
- before this, the tables did not have ids (at least from what I see in the code)
- you added ids to support hidden columns
- but now you say that `update_dropdowns_trips` is triggered only when we are inside trip table and will not affect any other table
- so why do you need the table id now? Why not just remove it and go back to the code without ids?

shankari:
@achasmita There is also a merge conflict; please resolve

pages/data.py
achasmita: For ```update_dropdowns_trips``` callback  ID is essential because it specifies the target output where the hidden columns will be updated. Without the ID, it would not know where to apply changes triggered by the callback.

shankari: @achasmita but with the current code, we have the same callback ID for all tables. So it still doesn't know which table to apply changes triggered by the callback.

achasmita: It will pass the hidden_columns property into populate_datatable only when we are in trips-table tab as the button is inside the trips table. To make it more clear I also have updated code  to pass table id only for trips table.

shankari (approved):
There are at least 4 extraneous whitespaces in this file. At this point, we have gone through multiple rounds of review feedback, and this is the closest we have come to a clean PR. So I am going to edit the file myself and merge. But @achasmita please submit high-quality, clean PRs in the future so that we don't need as much back and forth.

I also think that the admin dashboard data tabs could use some cleanup and refactoring. Maybe @JGreenlee can also make some suggestions when moving from the UI to the admin dashboard

shankari (approved):

shankari:
Squash-merging to address multiple back and forth changes. Please handle properly while merging.

PR #98, opened 2024-01-28
🔎  Exclude `test` Users Checkbox + filter from all data

JGreenlee:
(this PR extends on #96 since the changes are fairly intertwined)

### add filter to exclude 'test' users

'test' users are devs or program admininstrators whose data we usually want to keep separate from that of the participants themselves, identified by subgroup 'test' - in other words, if the opcode has "test" in it we should be able to filter out their data.

A checkbox is added below the datepicker to "Exclude 'test' users" (checked by default)

Filtering 'test' users from the UUIDs df is a simple task - drop any rows where 'user_token' contains 'test'.
But the other dfs (for confirmed trips, survey responses, trajectories) do not have the 'user_token' column - only 'user_id'.
So every time we want to filter these dfs, we'd have to cross-reference with the UUIDs table and lookup which 'user_id's correspond to test 'user_token's.
(I think of this like a 'join' in SQL)

I think a more performant solution is to keep a separate 'excluded uuids' store, which is just a plain list of excluded UUIDs that we determined while loading the 'uuids' store. The other stores can depend on this, so they will not need to depend on the entire 'uuids' store and do those cross-referencing lookups each time.

### DIFF
--- a/pages/data.py
+++ b/pages/data.py
@@ -37,9 +37,9 @@ def clean_location_data(df):
         df['data.end_loc.coordinates'] = df['data.end_loc.coordinates'].apply(lambda x: f'({x[0]}, {x[1]})')
     return df
 
-def update_store_trajectories(start_date: str, end_date: str):
+def update_store_trajectories(start_date: str, end_date: str, tz: str):
     global store_trajectories
-    df = query_trajectories(start_date, end_date)
+    df = query_trajectories(start_date, end_date, tz)
     records = df.to_dict("records")
     store = {
         "data": records,
@@ -58,9 +58,9 @@ def update_store_trajectories(start_date: str, end_date: str):
     Input('store-trajectories', 'data'),
     Input('date-picker', 'start_date'),
     Input('date-picker', 'end_date'),
-
+    Input('date-picker-timezone', 'value'),
 )
-def render_content(tab, store_uuids, store_trips, store_demographics, store_trajectories, start_date, end_date):
+def render_content(tab, store_uuids, store_trips, store_demographics, store_trajectories, start_date, end_date, timezone):
     data, columns, has_perm = None, [], False
     if tab == 'tab-uuids-datatable':
         data = store_uuids["data"]
@@ -98,7 +98,7 @@ def render_content(tab, store_uuids, store_trips, store_demographics, store_traj
         #Here we query for trajectory data once "Trajectories" tab is selected
         start_date, end_date = start_date[:10], end_date[:10] # dates as YYYY-MM-DD
         if store_trajectories == {}:
-            store_trajectories = update_store_trajectories(start_date, end_date)
+            store_trajectories = update_store_trajectories(start_date, end_date, timezone)
         data = store_trajectories["data"]
         if data:
             columns = list(data[0].keys())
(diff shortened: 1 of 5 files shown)


JGreenlee:
Here is how I have the checkbox currently implemented

image

I think it may be worthwhile to consider other designs/ layouts because we now have multiple options for global filtering, and it's starting to get cluttered

JGreenlee:
95ded215f73bde8332fa58d173613f34228629cc: organize filtering options into a collapsible menu

Filter menu closed
image

Filter menu open
image

pages/data.py
```diff
@@ -84,7 +85,7 @@ def render_content(tab, store_uuids, store_trips, store_demographics, store_traj
             data = list(data.values())[0]
             columns = list(data[0].keys())
         # for multiple survey, create subtabs for unique surveys
-        else:
+        elif len(data) > 1:
```
shankari, line 88: This makes sense, but shouldn't `len(data)` always be greater than 1? When would it not be?

shankari: Ah answering my own question: we could only have test users, in which case, this all demographic data would be filtered out if we are excluding them. This was not an issue earlier when if we had users, we would have demographic data.

utils/db_utils.py
```diff
@@ -15,6 +15,16 @@
 from utils import permissions as perm_utils
 from utils.datetime_utils import iso_range_to_ts_range
 
+def df_to_filtered_records(df, col_to_filter=None, vals_to_exclude: list[str] = []):
+    """
+    Returns a dictionary of df records, given a dataframe, a column to filter on,
+    and a list of values that rows in that column will be excluded if they match
+    """
+    if df.empty: return []
+    if col_to_filter and vals_to_exclude:
```
shankari, line 24: this should avoid the filtering operation when there are no vals to exclude.

```suggestion
    if col_to_filter and vals_to_exclude and len(vals_to_exclude) > 0:
```

shankari (changes requested):
Otherwise LGTM!

utils/db_utils.py
JGreenlee: I read that empty lists are falsy in Python, so I thought we wouldn't have to specifically check the length.

Do you think it's better to just be explicit?

pages/data.py
JGreenlee: It was **any** time there were 0 demographics responses, the table would stop working.

Presumably if there are any users, they will have recorded a demographics response. But what about ghost users or users whose demographics survey hasn't been pushed up yet?

So it was an pre-existing problem, but a bit of an edge case. With filtering added it was much more likely to occur, I noticed that and fixed it

utils/db_utils.py
shankari: I did not know that, but I checked and your reading was correct.

```
>>> t = []
>>> if t:
...     print("true")
...
>>> if not t:
...     print("false")
...
false
```

Interesting! That was non-obvious to me (at least), so I would suggest making it explicit or adding a comment.

JGreenlee: added comment

shankari (approved):

PR #99, opened 2024-01-30
Make QR codes visible on token page and delete tokens.

achasmita:
- QR codes are not visible in token table
- Regenerating the QR codes initially makes it visible on the table

### DIFF
--- a/pages/tokens.py
+++ b/pages/tokens.py
@@ -16,6 +16,7 @@ from utils.generate_qr_codes import saveAsQRCode
 from utils.generate_random_tokens import generateRandomTokensForProgram
 from utils.permissions import get_token_prefix, has_permission
 
+token_deletion_enabled = os.getenv('TOKEN_DELETION_ENABLED', 'False').lower() == 'true'
 
 if has_permission('token_generate'):
     register_page(__name__, path="/tokens")
@@ -94,7 +95,7 @@ layout = html.Div(
         #Create a confirmation dialog provider with a button(Delete Selected tokens) triggering a confirmation message.
         #Allow users to delete tokens with users confirmation.
         dcc.ConfirmDialogProvider(
-            children=html.Button('Delete Selected Tokens', id = 'delete-button', style = {'float': 'left',  
+            children=html.Button('Delete Selected Tokens', id = 'delete-button', disabled = not (os.getenv('TOKEN_DELETION_ENABLED', 'False').lower() == 'true'), style = {'float': 'left',  
             'font-size': '14px', 'width': '160px', 'display': 'block', 'margin-bottom': '10px',
             'margin-right': '5px', 'height':'40px', 'verticalAlign': 'top', 'background-color': 'green',
             'color': 'white',}),
@@ -131,7 +132,7 @@ def update_selected_rows(selected_rows):
 #Delete selected rows based on confirmation
 def delete_selected_rows(submit_clicks, current_data, selected_rows):
     #Check if the delete confirmation button is clicked
-    if submit_clicks:
+    if submit_clicks and token_deletion_enabled:
         #Remove selected rows from the current data
         current_data = [row for i, row in enumerate(current_data) if i not in selected_rows]
         df = query_tokens()
--- a/pages/tokens.py
+++ b/pages/tokens.py
@@ -3,6 +3,7 @@ import zipfile
 
 import pandas as pd
 
+import dash
 import dash_bootstrap_components as dbc
 from dash import dcc, html, Input, Output, callback, State, register_page, dash_table
 
@@ -96,7 +97,7 @@ layout = html.Div(
         ]),
 
         html.Div(id='token-table'),
-
+        html.Div(id='deleted-row-msg'),
         html.Br(),
         html.Button(children='Export QR codes', id='token-export', n_clicks=0, style={
             'font-size': '14px', 'width': '140px', 'display': 'block', 'margin-bottom': '10px',
@@ -106,6 +107,7 @@ layout = html.Div(
     ]
 )
 generate_qrcodes_for_all_tokens()
+
 
 @callback(
     Output('token-generate', 'n_clicks'),
@@ -127,6 +129,29 @@ def generate_tokens(n_clicks, program, token_length, token_count, out_format, ch
     tokens_table = populate_datatable()
     return 0, tokens_table
 
+
+@callback(
+    Output('deleted-row-msg', 'children'),
+    [Input('tokens-table', 'data_previous')],
+    [State('tokens-table', 'data')]
+)
+
+def delete_token_on_row_selection(previous, current):
+    if previous is None:
+        dash.exceptions.PreventUpdate()
+    else:
+        for row in previous:
+            if row not in current:
+                return delete_token([row['token']])
+
+
+def delete_token(token_list):
+    delete_list = [{"token":t} for t in token_list]
+    edb.get_token_db().delete_one(delete_list[0])
+    qrcode_file_path = os.path.join(QRCODE_PATH, f"{token_list[0]}.png")
+    if os.path.exists(qrcode_file_path):
+        os.remove(qrcode_file_path)
+    return 'Just removed {}'.format(token_list)
 
 @callback(
     Output('download-token', 'data'),
@@ -172,9 +197,9 @@ def populate_datatable():
         markdown_options={"html": True},
         style_table={'overflowX': 'auto'},
         export_format='csv',
+        row_deletable=True
     )
 
-def query_tokens():
-    query_result = edb.get_token_db().find({}, {"_id": 0})
-    df = pd.json_normalize(list(query_result))
-    return df
+
+
+
--- a/pages/tokens.py
+++ b/pages/tokens.py
@@ -30,7 +30,6 @@ def generate_qrcodes_for_all_tokens(): #generate QR codes for all tokens stored 
     if not os.listdir(QRCODE_PATH):
         df = query_tokens()
         for _, row in df.iterrows():
-            df = query_tokens()
             saveAsQRCode(QRCODE_PATH, row['token'])
 
 layout = html.Div(
--- a/pages/tokens.py
+++ b/pages/tokens.py
@@ -9,6 +9,8 @@ from dash import dcc, html, Input, Output, callback, State, register_page, dash_
 
 from emission.storage.decorations.token_queries import insert_many_tokens
 import emission.core.get_database as edb
+import base64
+import io
 
 from utils.generate_qr_codes import saveAsQRCode
 from utils.generate_random_tokens import generateRandomTokensForProgram
@@ -21,16 +23,6 @@ if has_permission('token_generate'):
 intro = """## Tokens"""
 QRCODE_PATH = 'assets/qrcodes'
 
-def query_tokens():
-    query_result = edb.get_token_db().find({}, {"_id": 0})
-    df = pd.json_normalize(list(query_result))
-    return df
-
-def generate_qrcodes_for_all_tokens(): #generate QR codes for all tokens stored in the database
-    if not os.listdir(QRCODE_PATH):
-        df = query_tokens()
-        for _, row in df.iterrows():
-            saveAsQRCode(QRCODE_PATH, row['token'])
 
 layout = html.Div(
     [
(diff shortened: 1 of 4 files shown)


shankari:
@achasmita this PR is still a draft
Where is the testing done?

achasmita:
I made use of dash ```dcc.ConfirmDialogProvider``` to create a confirmation dialog that appears when user select ```Delete Selected Rows``` button
>  https://dash.plotly.com/dash-core-components/confirmdialogprovider

- User can select tokens using checkbox on each rows:

Screen Shot 2024-02-02 at 10 29 51 AM

- And click ```Delete Selected Tokens```:
Screen Shot 2024-02-02 at 10 30 01 AM

- After clicking the  ```Delete Selected Tokens``` button, dialog box will pop up for user confirmation:
Screen Shot 2024-02-02 at 10 30 20 AM

If user select ok it will delete respective token else cancel the deletion.

shankari:
I still don't see "Testing done" for the initial fix (QR code not visible).

achasmita:
I have tested and QR code is visible and also tokens get deleted. I generated two new tokens and tried deleting them also verified it by checking it manually.
Screen Shot 2024-02-03 at 10 43 46 PM

shankari:
>  have tested and QR code is visible 

Screenshot please.

Also, while reporting command line outputs, it is better to copy and paste rather than include a screenshot since the screenshot is not searchable

achasmita:
Screen Shot 2024-02-04 at 7 58 30 PM

shankari:
@achasmita you want to not display real QR codes on tokens on a public website. I have now edited to remove them and only leave in the test tokens. Please be cognizant of what you put up publicly versus internally.

pages/tokens.py
```diff
@@ -84,17 +95,69 @@
...
+            message='Are you sure you want to delete selected token(s)?',
+        ),
         html.Button(children='Export QR codes', id='token-export', n_clicks=0, style={
             'font-size': '14px', 'width': '140px', 'display': 'block', 'margin-bottom': '10px',
             'margin-right': '5px', 'height':'40px', 'verticalAlign': 'top', 'background-color': 'green',
             'color': 'white',
         }),
     ]
 )
+generate_qrcodes_for_all_tokens() #Generate QR codes for all tokens stored in database
```
shankari, line 120: This calls `saveAsQRCode` on every token. So if we invoke this on every load, we are creating the QR code files multiple time and overriding them over and over.
- do we need to save the QR codes as files? We need to convert them to files before downloading, but it is not clear that storing them on an arbitrary filesystem is a great idea
- is this invoked on every login or every instance launch? If on every login, see note about extraneous work above.

shankari (changes requested):

pages/tokens.py
shankari: @achasmita I see that you have made changes related to the second point. However, I do not see your answer to the first question.

pages/tokens.py
```diff
@@ -161,9 +224,10 @@ def populate_datatable():
...
         style_table={'overflowX': 'auto'},
         export_format='csv',
+        row_selectable='multi', #Allow multiple row selection
+        selected_rows=[], #Initialize selected_rows as an empty list
     )
 
-def query_tokens():
-    query_result = edb.get_token_db().find({}, {"_id": 0})
-    df = pd.json_normalize(list(query_result))
-    return df
```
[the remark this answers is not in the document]
shankari: @achasmita this is not answered

pages/tokens.py
achasmita: I was exploring other option to generate QR codes dynamically and display them in the token table while allowing user to export QR code images as files. One of the options I found was converting QR code image to a base64 encoded string. That way, we can embed image directly into HTML without the need for a separate file. This will require the following changes to the code.
  - Need to update ```saveAsQRCode``` function to generate QR codes dynamically and return the QR code image as bytes
  - Need to modify ```populate-datatable``` function to generate QR codes dynamically for each token.
  - Need to update ```export_tokens``` callback to save dynamically generated QR code images as files when exporting.

shankari:
@achasmita why is this back in ready for review? I don't see changes other than the merge and the `QRCODE_WITH_PATH` directoy check, which will be obsolete once you switch to generating QR codes on the fly

shankari:
@achasmita as I suggested earlier, please make the token deletion functionality configurable. We currently do not have permission from our cyber team to modify the database. I don't anticipate any issues getting that permission, but we cannot enable the functionality until we have the permission. To make sure that we don't enable it inadvertently, please make sure that this is an env variable (similar to the COGNITO keys) instead of being specified deployment by deployment in the dynamic config.

achasmita:
Testing is done. The ```token_deletion_enabled``` is set false for now and delete button remains disable when it is false.

utils/generate_qr_codes.py
```diff
@@ -3,22 +3,25 @@
...
+def saveAsQRCode(token):
+    qrcode_data = f"nrelopenpath://login_token?token={token}"
+    qr = qrcode.QRCode(version=1, error_correction=qrcode.constants.ERROR_CORRECT_L, box_size=10, border=4)
+    qr.add_data(qrcode_data)
+    qr.make(fit=True)
+    img = qr.make_image(fill_color="black", back_color="white")
+    img_bytes = io.BytesIO()
+    img.save(img_bytes, format='PNG')
+    img_bytes.seek(0)
+    return img_bytes
```
JGreenlee, line 24: I see that we can generate the images using the `qrcode` library and not using `pil`

I don't see any other uses of `pil`/`pillow` in the codebase. So can we remove `pillow` from `requirements.txt`?

pages/tokens.py
```diff
@@ -122,24 +170,24 @@ def generate_tokens(n_clicks, program, token_length, token_count, out_format, ch
...
-                    zf.write(file_path, file_path[len_dir_path:])
-
+            for token in edb.get_token_db().find({}, {"token": 1}):
+                img_bytes = saveAsQRCode(token['token'])
+                zf.writestr(f"{token['token']}.png", img_bytes.getvalue())
+        bytes_io.seek(0)
+        return bytes_io.read()
     if n_clicks > 0:
-        return dcc.send_bytes(zip_directory, "tokens.zip")
+        return dcc.send_bytes(zip_directory(), "tokens.zip")
```
JGreenlee, line 182: It isn't clear to me why this was changed from `zip_directory` to `zip_directory()`. Can you clarify?

JGreenlee: Here is my interpretation, let me know if this is incorrect-

If you pass a plain function to `dcc.send_bytes`, it calls that function and provides a `bytes_io` as the argument to your function.
But instead of accepting that argument, we are now creating our own `bytes_io` and returning that.

utils/generate_qr_codes.py
```diff
@@ -3,22 +3,25 @@
 import base64
 import os
 import qrcode
-import PIL as pil
+# import PIL as pil
```
JGreenlee, line 6: If we can confirm that pillow is not needed, this can be removed too

JGreenlee (commented):
This looks really clean! I have left a few questions below

utils/generate_qr_codes.py
achasmita: Yes, ```pil``` was used before to draw ```qrcode```, but now we are not using it anymore. I will remove it.

pages/tokens.py
achasmita: When we pass a function to ```dcc.send_bytes```, Dash calls that function and expects it to return bytes directly. In this case, ```zip_directory``` is a function that returns bytes, so using ```zip_directory()``` correctly calls the function and provides the bytes directly to dcc.send_bytes. This ensures that the bytes of the zipped directory are correctly sent for download as a zip file.

pages/tokens.py
```diff
@@ -84,10 +87,21 @@
...
+        #Create a confirmation dialog provider with a button(Delete Selected tokens) triggering a confirmation message.
+        #Allow users to delete tokens with users confirmation.
+        dcc.ConfirmDialogProvider(
+            children=html.Button('Delete Selected Tokens', id = 'delete-button', disabled = not (os.getenv('TOKEN_DELETION_ENABLED', 'False').lower() == 'true'), style = {'float': 'left',  
+            'font-size': '14px', 'width': '160px', 'display': 'block', 'margin-bottom': '10px',
+            'margin-right': '5px', 'height':'40px', 'verticalAlign': 'top', 'background-color': 'green',
+            'color': 'white',}),
+            id='confirm-delete',
+            message='Are you sure you want to delete selected token(s)?',
+        ),
```
jiji14, line 104: Can we add styles when the button is disabled? For example, change the background color to grey when it is disabled and change the background color to green when it is enabled. When I tested it, I got confused because I thought the button was clickable, but it was not.

Also, for testing purposes, I set `TOKEN_DELETION_ENABLED` to `True`, but the button was still not clickable. Is there any other setting I need to change to enable the delete function?

achasmita: Yes, that's a great idea. Styling the active and disabled button differently easily help user differentiate between the two states.

You also need to make this true
``` token_deletion_enabled = os.getenv('TOKEN_DELETION_ENABLED', 'False').lower() == 'true'```

Comment (shankari) [reply pages/tokens.py:104]: @achasmita @JGreenlee I don't understand this comment

> You also need to make this true `os.getenv('TOKEN_DELETION_ENABLED', 'False').lower()`

If @jiji14 sets `TOKEN_DELETION_ENABLED` to True, then `os.getenv` will return `True` instead of the default `False` so it should still work, right?

File: pages/tokens.py [hunk 7]
```diff
@@ -96,6 +110,42 @@
...
+    State('tokens-table', 'data'),
+    State('selected-rows', 'data'),
+    prevent_initial_call=True
+)
+#Delete selected rows based on confirmation
+def delete_selected_rows(submit_clicks, current_data, selected_rows):
+    #Check if the delete confirmation button is clicked
+    if submit_clicks and token_deletion_enabled:
+        #Remove selected rows from the current data
+        current_data = [row for i, row in enumerate(current_data) if i not in selected_rows]
```
shankari, line 137: cleanup fix: I would suggest re-reading the table from the database instead of making local changes here. The admin dashboard runs on the server, so we do not anticipate network drops between the python code and the DB. And this will ensure that if there are issues with deleting the data, we don't show a wrong view to the user. It will also handle consistency between parallel sessions (e.g. another admin is creating tokens at the same time).

If we do want to keep this code (which I don't recommend), we have to make the local changes after the DB changes, and use a try/catch block to ensure that the local changes only happen if the DB changes were successful.

utils/generate_qr_codes.py
```diff
@@ -3,22 +3,25 @@
...
     tokens = []
     with open(filename) as fp:
         tokens = [t.strip() for t in fp.readlines()]
     return tokens
 
-def saveAsQRCode(outdir, token):
-    qrcode_data = "nrelopenpath://login_token?token="+token
-    qrcode_img = qrcode.make(qrcode_data)
-    draw = pil.ImageDraw.Draw(qrcode_img)
-    draw.text((55,10), token, fill=0, align="center", anchor="mm")
```
shankari, line 18: this text version of the opcode is missing from the newly rewritten value. We should retain this functionality so that admins can see which QR code they are handing out without having to use a QR code scanner, which may not display the full opcode anyway if it is long. Note that this may require re-adding PIL

pages/tokens.py
```diff
@@ -161,9 +209,13 @@ def populate_datatable():
...
+        row_selectable='multi', #Allow multiple row selection
+        selected_rows=[], #Initialize selected_rows as an empty list
     )
 
 def query_tokens():
     query_result = edb.get_token_db().find({}, {"_id": 0})
     df = pd.json_normalize(list(query_result))
     return df
+
+
```
shankari, line 221: Extraneous whitespace

pages/tokens.py
shankari: @jiji14 have you done this styling yet?

pages/tokens.py
```diff
@@ -3,22 +3,25 @@
...
+import base64
+import io
 
 from utils.generate_qr_codes import saveAsQRCode
 from utils.generate_random_tokens import generateRandomTokensForProgram
 from utils.permissions import get_token_prefix, has_permission
 
+token_deletion_enabled = os.getenv('TOKEN_DELETION_ENABLED', 'False').lower() == 'true'
 
 if has_permission('token_generate'):
```
shankari, line 21: we should also show the token page only when tokens are not autogenerated. If tokens are autogenerated, they don't need to be stored in the token_db and don't need to be added or deleted.

shankari (changes requested):
@jiji14 there are a couple of high level changes that need to be fixed before this can be merged. We may need to add pillow back!

PR #103, opened 2024-02-20
Loadi UUIDs  datatable with only 25 user_id initially.

achasmita:
- loads only data for first 25 user_id initially
- Allow user to load more data when available  by selecting 'load more data' button
- All the columns in UUIDs datatable will be stored initially instead after clicking data-page or UUIDs tab

### DIFF
--- a/pages/data.py
+++ b/pages/data.py
@@ -54,7 +54,7 @@ def update_store_trajectories(start_date: str, end_date: str, tz: str):
 @callback(
     Output('tabs-content', 'children'),
     Input('tabs-datatable', 'value'),
-    Input('store-uuids', 'data'),
+    Input('store-user-stats', 'data'),
     Input('store-trips', 'data'),
     Input('store-demographics', 'data'),
     Input('store-trajectories', 'data'),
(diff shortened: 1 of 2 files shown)


pages/data.py
```diff
@@ -176,6 +188,24 @@ def update_dropdowns_trips(n_clicks, button_label):
...
+)
+#Load more uuid_table data if available.
+def update_uuids(n_clicks, store_uuids, current_data):
+    if n_clicks:
+        if len(current_data) >= 25:
+            additional_data = db_utils.add_user_stats(store_uuids["data"][25:])
+            current_data.extend(additional_data)
+            df = pd.DataFrame(current_data)
+            return df.to_dict('records'), {'display': 'none'}
+        return {}
```
jiji14, line 207: I don't think you need to use nested conditions here!
```python
    if n_clicks and len(current_data) >= 25:
        additional_data = db_utils.add_user_stats(store_uuids["data"][25:])
        current_data.extend(additional_data)
        df = pd.DataFrame(current_data)
        return df.to_dict('records'), {'display': 'none'}
    return {}
```

pages/data.py
```diff
@@ -176,6 +188,24 @@ def update_dropdowns_trips(n_clicks, button_label):
     return hidden_col, button_label
 
 
+@callback(
+    Output('uuids-table', 'data'), # Output all data in the uuid-table
+    Output('load-data', 'style'),
+    Input('load-data', 'n_clicks'), #number of clicks on the button
+    Input('store-uuids', 'data'),
+    State('uuids-table', 'data'), #State representing the current data on table
+    prevent_initial_call=True
```
jiji14, line 197: Can you please explain what `prevent_initial_call` does?

achasmita: It prevents callback from being triggered when the app is first loaded.

PR #106, opened 2024-03-20
[Bug fix] Trips trend on Home not working

jiji14:
### Overview
After our recent updates, the trip trends on the Home tab aren't working. The error occurs when the column data is converted to `datetime` with a specific format (ISO8601). 

#### [ Before ]
Screenshot 2024-03-20 at 4 11 15 PM

#### [ After ]
Screenshot 2024-03-20 at 4 12 30 PM

### DIFF
--- a/pages/home.py
+++ b/pages/home.py
@@ -64,7 +64,7 @@ def compute_sign_up_trend(uuid_df):
 
 
 def compute_trips_trend(trips_df, date_col):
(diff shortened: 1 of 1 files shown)


jiji14:
@JGreenlee  Can you please confirm that this change doesn't affect your update?

shankari:
@JGreenlee tagging you for review

JGreenlee (approved):
LGTM

I was not previously aware of this issues, so it does not affect anything I was working on for admin dashboard. Thanks for finding the bug and fixing promptly!

PR #107, opened 2024-03-22
Removes pillow from requirements.txt

achasmita:
Removes pillow from requirements.txt


shankari:
This change is fairly simple, but it also is very messy and has dependencies on other changes that are waiting to be cleaned up (QR code changes). Closing this so that we can fix it properly as part of the admin dashboard changes @JGreenlee  for visibility

I think that once the QR code generation changes (which we haven't finished yet), are done, we can potentially remove pillow, which is a frequent source of vulnerabilities.

PR #108, opened 2024-03-22
Integrates PygWalker to visualize data

achasmita:


achasmita:
I have created a simple dropdown for now to switch between user data and trips data, we can still more data if needed.
- It allows to select option from dropdown and load data on filelds based on option selected

Screen Shot 2024-03-22 at 5 11 33 PM
Screen Shot 2024-03-22 at 5 12 01 PM

TeachMeTW:
@JGreenlee I am moving this to #130 since I cannot fork or edit this branch; tried to do so but says I do not have permission. This can be closed.

Issue #110, opened 2024-03-23
🚑 Maps not visible on the staging admin dashboard


shankari:
https://github.com/e-mission/op-admin-dashboard/assets/2423263/b9c8748d-60d0-463e-8598-9bb364846377

shankari:
Error from the logs:

```
ERROR:app_sidebar_collapsible:Exception on /_dash-update-component [POST]
Traceback (most recent call last):
  File "/root/miniconda-23.5.2/envs/emission/lib/python3.9/site-packages/flask/app.py", line 2529, in wsgi_app
    response = self.full_dispatch_request()
  File "/root/miniconda-23.5.2/envs/emission/lib/python3.9/site-packages/flask/app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/root/miniconda-23.5.2/envs/emission/lib/python3.9/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "/root/miniconda-23.5.2/envs/emission/lib/python3.9/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/root/miniconda-23.5.2/envs/emission/lib/python3.9/site-packages/dash/dash.py", line 1310, in dispatch
    ctx.run(
  File "/root/miniconda-23.5.2/envs/emission/lib/python3.9/site-packages/dash/_callback.py", line 442, in add_context
    output_value = func(*func_args, **func_kwargs)  # %% callback invoked %%
  File "/usr/src/app/pages/map.py", line 239, in update_user_ids_options
    user_ids_options, user_ids = create_user_ids_options(trips_data['users_data_by_user_id'])
KeyError: 'users_data_by_user_id'
```

shankari:
@jiji14 @achasmita @JGreenlee for visibility

PR #112, opened 2024-03-26
[Bug Fix] Maps not visible on the staging admin dashboard

jiji14:
[ Related Issue ]
https://github.com/e-mission/op-admin-dashboard/issues/110

[ Overview ]
Based on error log, the issue comes from `KeyError: 'users_data_by_user_id'`, so I added error handling for this 

Screenshot 2024-03-26 at 9 23 08 AM

### DIFF
--- a/pages/map.py
+++ b/pages/map.py
@@ -283,7 +283,6 @@ def update_output(map_type, selected_user_ids, selected_user_emails, selected_us
     user_ids = set(selected_user_ids) if selected_user_ids is not None else set()
     user_modes=set(selected_user_modes) if selected_user_modes is not None else set()
     coordinates = get_map_coordinates(trips_data.get('users_data_by_user_mode', {}), user_modes)
-
     if selected_user_emails is not None:
         for user_email in selected_user_emails:
             user_ids.add(str(ecwu.User.fromEmail(user_email).uuid))
--- a/pages/map.py
+++ b/pages/map.py
@@ -22,7 +22,6 @@ register_page(__name__, path="/map")
 
 intro = """## Map"""
 
-
 def create_lines_map(trips_group_by_user_id, user_id_list):
     start_lon, start_lat = 0, 0
     traces = []
@@ -236,7 +235,10 @@ layout = html.Div(
     Input('user-id-dropdown', 'value'),
 )
 def update_user_ids_options(trips_data, selected_user_ids):
-    user_ids_options, user_ids = create_user_ids_options(trips_data['users_data_by_user_id'])
+    user_ids_options, user_ids = list(), set()
+    if 'users_data_by_user_id' in trips_data:
+        user_ids_options, user_ids = create_user_ids_options(trips_data['users_data_by_user_id'])
+
     if selected_user_ids is not None:
         selected_user_ids = [user_id for user_id in selected_user_ids if user_id in user_ids]
     return user_ids_options, selected_user_ids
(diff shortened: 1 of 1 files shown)


shankari:
@jiji14 it seems like "works" here means that there is no error. But there are also no trips. Are you sure that there are no trips in the database? What does the trip trends table show? If you select a different date range, does the map show up?

Concretely, on staging (which is where we first encountered this), I do see trips in the trips trend table. So why are is the `users_data_by_user_id` table not created correctly.

Screenshot 2024-03-26 at 11 53 09 AM

jiji14:
@shankari 
Thank you for your comment. Yes, I can see the trips in the database. Now I am investigating why the `users_data_by_user_id` table was not created correctly.

jiji14:
@shankari 

There was a dataset where 'data.user_input.mode_confirm' did not exist in the column, causing the data to be missing on the map. I'm not sure if this case only exists in the development environment, but I labeled that trip as 'unlabeled' so it could show up on the map. When I shared the map page during the meeting, the trip data between 12/1/2023 and 12/2/2023 did not show up, but you can now see that it shows up as an unlabeled trip in screenshot.

Screenshot 2024-03-26 at 4 36 07 PM
Screenshot 2024-03-26 at 4 36 19 PM

shankari:
> There was a dataset where 'data.user_input.mode_confirm' did not exist in the column, causing the data to be missing on the map.

Note that the admin dashboard does not work with static datasets, but instead, with the data that is coming in from the phones real-time. We sometimes work with snapshots of that real-time data, but this is not static analysis.

> I'm not sure if this case only exists in the development environment, but I labeled that trip as 'unlabeled' so it could show up on the map. 

If you are using a snapshot of real data, I am not sure why it would happen only in the development environment. This can happen any time none of the selected trips are labeled. For the snapshot you loaded, you can check the "Data" tab - and see if the mode_confirm is filled in. If it is not, this is clearly the problem.

shankari:
I can verify that the current staging does not have any `mode_confirm` column because none of the trips were labeled. I suspect that this is the issue, and I am moving on to review this change now.
Screenshot 2024-03-26 at 5 43 10 PM

jiji14:
@shankari 
I thought we used dummy datasets for dev environment, not the real datasets. I got your point. Thanks for clarification. I also checked the "Data" tab doesn't not have `mode_confirm` column

Screenshot 2024-03-26 at 5 51 20 PM

pages/map.py
```diff
@@ -236,7 +238,10 @@ def create_user_modes_options(trips_group_by_user_mode):
     Input('user-id-dropdown', 'value'),
 )
 def update_user_ids_options(trips_data, selected_user_ids):
-    user_ids_options, user_ids = create_user_ids_options(trips_data['users_data_by_user_id'])
+    user_ids_options, user_ids = list(), set()
+    if 'users_data_by_user_id' in trips_data:
+        user_ids_options, user_ids = create_user_ids_options(trips_data['users_data_by_user_id'])
+
```
shankari, line 244: given that the issue is with the `mode_confirm` column being missing, do you even need these other checks?

Note that if `users_data_by_user_id` is not in trips_data, we return empty values, which may hide other errors (like the one with the missing mode_confirm)

shankari (commented):
@jiji14 can we clean up this PR before I merge?

pages/map.py
jiji14: @shankari 
Cleaned it and checked the local is running well!

What I understand is, if `users_data_by_user_id` is not present in `trips_data`, we need to address this issue. So this line might obscure certain error cases.

shankari (approved):

shankari:
@jiji14 I am squash-merging this to avoid the churn with the commits + reverts. Please make sure to account for that while starting your next PR!

Issue #43, opened 2023-05-22
:lock: load the OpenPATH logo from the local filesystem


shankari:
Loading it from the NREL website is bad because:
- it introduces a cross-site dependency
- it introduces extra cookies from the NREL site that are not required for our codebase

I have removed the OpenPATH logo for now; we should reintroduce it, load it from the local filesystem and add a favicon

PR #113, opened 2024-03-26
Redesign Build Deployment Process (External)

MukuFlash03:
This PR serves as the implementation for [this issue](https://github.com/e-mission/e-mission-docs/issues/1048) for redesigning our build and deployment processes across e-mission-server, join, admin-dash, public-dash primarily.

Collaborating on this along with @nataliejschultz .

All four redesign PRs for the external repos:
E-mission-server: https://github.com/e-mission/e-mission-server/pull/961
Join: https://github.com/e-mission/nrel-openpath-join-page/pull/29
Admin-dash: https://github.com/e-mission/op-admin-dashboard/pull/113
Public-dash: https://github.com/e-mission/em-public-dashboard/pull/125

-------------

A rough plan can be:

1st phase
- Consolidating Differences
  - Minimize the differences in internal and external versions, cleanup unnecessary files, so on.
- Image_Build_Push.yml: 
  - CI / CD via GitHub actions to build and push images to Dockerhub for all four repos based on the current process for e-mission-server.
  - Image build push only builds for that specific repo when merge is made to a specific branch.
  - Additionally, for e-mission-server, image-build-push will now also need to trigger the GitHub action in join, admin-dash, public-dash.

2nd phase: 
- Multi-tier uniform repo structure in nrelopenpath
  - Need to change the internal repos to pull from the externally built images for each repo without duplicating the entire external code repository internally.


MukuFlash03:
Checking in the first initial set of changes for consolidating the differences in the external and internal versions.
These are mostly related to config files based on the [differences identified here](https://github.com/e-mission/e-mission-docs/issues/1048#issuecomment-1998477929)

Currently included these changes:

1. Image_build_push.yml
- Involves the image_build_push GitHub action automation to build and push images to Dockerhub whenever changes are merged to the specified branch.
- I have used a local test branch (image-push-merge) which can be changed to master once changes are finalized.

---------

Next are various changes to consolidate differences; made in [this commit](https://github.com/e-mission/op-admin-dashboard/commit/a24bd99cb934308a1e33f1f9859af41810044738).

2. Loading OpenPATH logo from local image rather than URL link.
- Changed to use a static local image to load the logo used in app_sidebar_collapsible.py

--------

3. Changed sed to jq in start.sh.
- Making it consistent to the internal repo version.

--------

3. Changed dash version to latest v 2.16.1
Detailed information available in the differences table mentioned at the top.
- Upgraded to the latest version so that we no longer need the workaround which involves using INDEX_STRING_NO_META. 
- Removed all uses of INDEX_STRING_NO_META.

---------

4. Modified Dockerfile(s)
- Removed assets/ from COPY command as only contents of assets/ were being copied to root app directory.
- In COPY assets command, copy entire directory instead of just css file.

nataliejschultz (commented):
Adding comments.

docker/Dockerfile
```diff
@@ -17,10 +19,10 @@ COPY ./pages ./
 WORKDIR /usr/src/app/utils
 COPY ./utils ./
 WORKDIR /usr/src/app
-COPY app.py config.py app_sidebar_collapsible.py assets globals.py globalsUpdater.py Procfile ./
+COPY app.py app_sidebar_collapsible.py globals.py globalsUpdater.py Procfile ./
 
 WORKDIR /usr/src/app/assets
-COPY assets/style.css ./
+COPY assets/ ./
```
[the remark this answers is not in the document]
MukuFlash03: Yes, I decided to change it so the assets/ directory is copied over correctly, which wasn't the case earlier.

I came across this issue when trying to read in the `openpath logo image` because it was being copied in root app directory and not in `assets` due to the `assets` directory being there in the `COPY` layer. 

The reason is that if any folder is mentioned in the COPY layer in a Dockerfile, it copies contents of the directory (and not the directory) in the root app directory.
Check the comment by user Claudiu [here](https://stackoverflow.com/a/30316600).

-----

Thus in the existing code too, the `COPY app.py assets...` layer was copying contents of assets, i.e. `style.css` to the working directory set before the command - `/usr/src/app`.
Hence, to retain assets/ directory structure which contains style.css, another COPY layer was used to copy style.css separately.

To fix this duplication, I removed assets from the first `COPY` layer and mentioned only `assets/` in the 2nd `COPY` layer. This resulted in maintaining the directory structure of assets and did not result in copying contents of `assets/ to /usr/src/app.`

MukuFlash03:
### Testing admin-dashboard observations

Got steps from @jiji14 which are applicable on existing code on main branch (before redesign code changes).
For prod like testing, use docker-compose-prod.yml file instead of docker-compose-dev.yml in the below commands.

```
# Copy config file
cp config-fake.py config.py

# Run build script 
docker compose -f docker-compose-dev.yml build

# Load data 
./docker/load_mongodump.sh  tar_data_file (name of the file containing data)

# Change line 10 in start.sh  
sed "s-localhost-${DB_HOST}-" conf/storage/db.conf.sample > conf/storage/db.conf

# Change DB_HOST in docker-compose-dev.yml 
DB_HOST=mongodb://db/openpath_stage

# Run docker containers
docker compose -f docker-compose-dev.yml up

# Go to localhost:8050 and login 
username  = hello  
password  = world
```

Steps after redesign
No need to copy config file or modify sed command as these are not used anymore in redesigned code changes.

```
# Run build script 
docker compose -f docker-compose-dev.yml build

# Load data 
./docker/load_mongodump.sh  tar_data_file (name of the file containing data)

# Change DB_HOST in docker-compose-dev.yml 
DB_HOST=mongodb://db/openpath_stage

# Run docker containers
docker compose -f docker-compose-dev.yml up

# Go to localhost:8050 and login 
username  = hello  
password  = world
```

MukuFlash03:
Tested external repo code both with redesign code changes.
Admin-dash working with data and values can be seen in dashboard.

Screenshots showing values being loaded

Data tab
admin-dash-ext-data

Tokens tab
admin-dash-ext-tokens

nataliejschultz:
Testing build from Mukul's server code:

Changed the build image in the Dockerfile to the consolidate-differences image:
Screenshot 2024-04-25 at 7 47 10 PM

Ran `$ docker build -t mukul-admindash .` to build the image.

Had to turn off the VPN momentarily due to an HTTP request being blocked + keeping it from building.

Loaded cortezebikes snapshot tar.gz data into the database using:

 `./load_mongodump.sh ../data/openpath_prod_cortezebikes_*_*.tar.gz`
I did not want to completely rebuild a db container and rename, so I changed the name of the database in load_mongodump.sh to just be db:

Screenshot 2024-04-25 at 10 51 45 PM

 Confirmed that the data was loaded in by using `docker exec -it db /bin/bash` and running:
```
mongo
show dbs
```
The data that I loaded was there:
Screenshot 2024-04-25 at 10 54 00 PM

Ran docker run:
```
$ docker run --name mukul-int-admin-dash-1 -d -e DASH_SERVER_PORT=8050 -e DB_HOST=mongodb://db/openpath_prod_cortezebikes -e WEB_SERVER_HOST=0.0.0.0 -e CONFIG_PATH="https://raw.githubusercontent.com/e-mission/nrel-openpath-deploy-configs/main/configs/" -e STUDY_CONFIG=cortezebikes -e DASH_SILENCE_ROUTES_LOGGING=False -e SERVER_BRANCH=master -e REACT_VERSION="18.2.0" -e AUTH_TYPE="basic" -p 8050:8050 --network emission mukul-admindash
```
Of note: I kept getting an error when the STUDY_CONFIG variable was enclosed in quotes. This is strange, because the CONFIG_PATH variable was fine to be enclosed in quotes, and the CONFIG_PATH is concatenated with STUDY_CONFIG after being read in as environment variables. 

After waiting for a moment for the data to load, I was able to access the login screen at localhost:8050. I logged in with hello:world, and saw that the data were present!

Screenshot 2024-04-25 at 11 01 05 PM

(intentionally sharing the least consequential columns)

The sign-ups trend graph was also shown, I just don't want to show any actual data. I have the screenshot to show it if necessary.

nataliejschultz:
Internally testing admin dash build: 

Cd’ed into local version of internal nrelopenpath multi-tier branch . Changed the Dockerfile to build off of the local image I created in the last test:

Screenshot 2024-04-26 at 12 55 00 AM


Built the image with

 ` docker build -t mukul-internal-admin-dash ./admin_dashboard`

Ran the container with:

```
docker run --name mukul-int-admin-dash-2 -d -e DASH_SERVER_PORT=8050 -e DB_HOST=mongodb://db/openpath_prod_cortezebikes -e WEB_SERVER_HOST=0.0.0.0 -e CONFIG_PATH="https://raw.githubusercontent.com/e-mission/nrel-openpath-deploy-configs/main/configs/" -e STUDY_CONFIG=cortezebikes -e DASH_SILENCE_ROUTES_LOGGING=False -e SERVER_BRANCH=master -e REACT_VERSION="18.2.0" -e AUTH_TYPE="basic" -p 8050:8050 --network emission mukul-internal-admin-dash
```
Note that I again had to remove the quotations around the value for STUDY_CONFIG.
After going to localhost:8050, everything looked exactly the same as before!
Screenshot 2024-04-26 at 12 53 12 AM
Screenshot 2024-04-26 at 12 49 57 AM



User numbers blocked out to protect user privacy.

Dockerfile
```diff
@@ -0,0 +1,39 @@
+# Please change once all PR changes are final, so it reads from shankari/e-mission-server
```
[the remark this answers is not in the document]
shankari: @nataliejschultz what are the path complications?

app_sidebar_collapsible.py
```diff
@@ -29,15 +30,17 @@
 import flask_talisman as flt
 
 
-
-OPENPATH_LOGO = "https://www.nrel.gov/transportation/assets/images/openpath-logo.jpg"
+OPENPATH_LOGO = os.path.join(os.getcwd(), "assets/openpath-logo.jpg")
+encoded_image = base64.b64encode(open(OPENPATH_LOGO, 'rb').read()).decode("utf-8")
```
shankari, line 34: This is clearly correct, but just checking on why we didn't want to keep using the public openpath logo file?

utils/cognito_utils.py
```diff
@@ -4,16 +4,16 @@
...
 
 
 def get_tokens(code):
-    client_id = CognitoConfig.CLIENT_ID
-    client_secret = CognitoConfig.CLIENT_SECRET
-    redirect_uri = CognitoConfig.REDIRECT_URL
-    token_endpoint = CognitoConfig.TOKEN_ENDPOINT
+    client_id = os.getenv("COGNITO_CLIENT_ID", '')
+    client_secret = os.getenv("COGNITO_CLIENT_SECRET", '')
+    redirect_uri = os.getenv("COGNITO_REDIRECT_URL", '')
```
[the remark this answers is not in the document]
shankari: or create a .env file as a template?

shankari (changes requested):
I have some comments, but this looks good in general. As soon as the images are changed from the ones in mukul's directory to the ones in mine, and these questions are answered, I am happy to merge

Dockerfile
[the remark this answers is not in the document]
nataliejschultz: @shankari @MukuFlash03 had run into some problems with the path in relation to copying the files in the COPY commands, and executing the dockerfile from the main directory solved the problem

app_sidebar_collapsible.py
nataliejschultz: Per your commit [comment here ](https://github.nrel.gov/nrel-cloud-computing/nrelopenpath-admin-dashboard/commit/47f023dc1d6b5a6531693a90e0575830691356ec)and the issue opened [here](https://github.com/e-mission/op-admin-dashboard/issues/43#issue-1718938898):

> Loading [the logo] from the NREL website is bad because:
> 
> * it introduces a cross-site dependency
> * it introduces extra cookies from the NREL site that are not required for our codebase
> 
> I have removed the OpenPATH logo for now; we should reintroduce it, load it from the local filesystem and add a favicon

Hence, that issue can be closed upon this merge 😄

utils/cognito_utils.py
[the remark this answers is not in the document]
nataliejschultz: [.env file](https://github.com/e-mission/op-admin-dashboard/blob/9f960446a84386734e7512aa4ae6842ab825828c/.env) added

nataliejschultz:
Merged changes from the tag automation branch, so now going to cleanup + prepare for merge with main/implementation

nataliejschultz:
To respond to [this](https://github.com/e-mission/op-admin-dashboard/pull/113#discussion_r1588739850) we ran into issues with the COPY function in the dockerfile. We can keep it as is, or I can change all of the docker composes to build off of the Dockerfile in the main directory. The two docker files are identical aside from their locations.

Other than that, I have merged the tag update functionality + prepared all of the files to access the correct branches and directories for the automated tag updating.

Dockerfile
shankari, line 1: why do we need this file? There is already a Dockerfile at `docker/Dockerfile`

requirements.txt
shankari, line 1: why are these changes needed?

utils/cognito_utils.py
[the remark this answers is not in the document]
shankari: The `.env` does not have the full list of environment variables

docker/start.sh
```diff
@@ -5,9 +5,8 @@ source setup/activate.sh
 echo "DB host = "${DB_HOST}
 if [ -z ${DB_HOST} ] ; then
     local_host=`hostname -i`
-    sed "s_localhost_${local_host}_" conf/storage/db.conf.sample > conf/storage/db.conf
-else
-    sed "s_localhost_${DB_HOST}_" conf/storage/db.conf.sample > conf/storage/db.conf
+    export DB_HOST=$local_host
+    echo "Setting db host environment variable to localhost"
```
shankari, line 9: This can be changed to be consistent with the start script changes in the server.

shankari (changes requested):

Dockerfile
nataliejschultz: This was from [here](https://github.com/e-mission/op-admin-dashboard/pull/113#discussion_r1589764463), but since we're using docker compose which builds a different Dockerfile, it can be removed

requirements.txt
nataliejschultz: From [here](https://github.com/e-mission/e-mission-docs/issues/1048#issuecomment-1998477929), on the addition of INDEX_STRING_NO_META:
> We searched for history behind this addition and found that it was done as a workaround to handle a security issue with a script injection bug.
We found that Shankari had filed an issue with the dash library repository and was able to elicit a response and a fix from the official maintainers of the library.
With the dash library version that fixed it 2.10.0, flask version <= 2.2.3 is needed.
Hence choosing next higher versions (2.14.1 which increased flask version limit), 2.14.2 (mentioned by a developer that it works).
Working with 2.16.1 latest version as well.
The issue no longer appears when tested with versions: 2.14.1, 2.14.2, 2.16.1.

.github/workflows/image_build_push.yml
```diff
@@ -0,0 +1,172 @@
...
+        if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
+          echo "Workflow_dispatch: New server image built and pushed, Updating image tag in .env"
+          echo "SERVER_IMAGE_TAG=$DOCKER_IMAGE_TAG_2" > .env
+          echo "COGNITO_CLIENT_ID=''" > .env
+          echo "COGNITO_CLIENT_SECRET=''" > .env
+          echo "COGNITO_REDIRECT_URL=''" > .env
+          echo "COGNITO_TOKEN_ENDPOINT=''" > .env
+          echo "COGNITO_USER_POOL_ID=''" > .env
+          echo "COGNITO_REGION=''" > .env
+          echo "COGNITO_AUTH_URL=''" > .env
```
shankari, line 108: I am confused. If the goal is to have all the environment variables stored in `.env`, then why are you overriding them here? And if the goal is not have them in here, then why do we need to fill them in?

Basically, if you are going to override the values with empty strings anyway, then it is not clear that you need the variables in the first place.

.github/workflows/image_build_push.yml
```diff
@@ -0,0 +1,172 @@
...
+        fi
+
+    - name: Add, Commit, Push changes to .env file
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
```
shankari, line 124: nit: make name more meaningful.

shankari (changes requested):
In addition to the main change below, you also need to copy over the changes from 
https://github.com/e-mission/em-public-dashboard/pull/125
in here

.github/workflows/image_build_push.yml
MukuFlash03: @nataliejschultz Had added these to the .env file in [this commit](https://github.com/MukuFlash03/op-admin-dashboard/commit/a93760165caae8c36857b5d6fc4d3dc70a5ce8cb).

The reason for doing this was that GitHub Actions was overwriting the .env file as a part of the workflow. During the workflow only the latest server tag is written into the .env file.

MukuFlash03: But I do see a problem with this:
- If we make Cognito variables as empty strings in workflow, original stored values will be lost.

Initially, when we built the logic for the CI/CD to update the server image tag, the .env file was meant to have just the server image tag (related commits: [1](https://github.com/MukuFlash03/op-admin-dashboard/commit/d98f75c8ba36a8b69c547f30e8840b6af773c7a5), [2](https://github.com/MukuFlash03/op-admin-dashboard/commit/d03ce01665757a97576920db600c89c18d0b3f51)).

MukuFlash03: Added a separate template file in [this commit](https://github.com/e-mission/op-admin-dashboard/pull/113/commits/3f66ef344a70aebf11913b6202a007b051f226c1).
Keeping the .env file only for latest server image tag.

shankari: @MukuFlash03 Are you sure that multiple `.env` files are supported in docker? How did you test this to make sure that if the values are set, they are visible in the codebase?

.env.cognito.template
shankari, line 1: @MukuFlash03 Are you sure that multiple .env files are supported in docker? How did you test this to make sure that if the values are set, they are visible in the codebase?

shankari (changes requested):
Also, please make sure to check that the file names are valid, consistent with https://github.com/e-mission/em-public-dashboard/pull/125#issuecomment-2289658512
and
https://github.com/e-mission/em-public-dashboard/pull/125#issuecomment-2289672873

.github/workflows/image_build_push.yml
MukuFlash03: I don't think they would be set directly.

I thought of this approach, I had mentioned this in the commit:
```
I am expecting the users would either set them directly in the system or use the docker compose prod yml file.
```
So, with this option, the cognito env file would literally just be a template.

---

The second option is that we will need to load the env variables from the specified file using the python library [python-dotenv](https://pypi.org/project/python-dotenv/). Will need to add this in requirements.txt.

So, the code in utils/cognito_utils.py (for instance) would look like:
```
from dotenv import load_dotenv

load_dotenv(.env.cognito)

def get_tokens(code):
    client_id = os.getenv("COGNITO_CLIENT_ID", '')
    client_secret = os.getenv("COGNITO_CLIENT_SECRET", '')
    redirect_uri = os.getenv("COGNITO_REDIRECT_URL", '')
    token_endpoint = os.getenv("COGNITO_TOKEN_ENDPOINT", '')

```

shankari: `python-dotenv` is overkill
I guess it is fair to assume that the users will set the values in the environment, we do assume that for the server code. Once we figure out how to maintain the list of environment variables from the server, we should use the same approach here. Ideally this would be through a search so that we don't have to maintain a separate template file and ensure that it doesn't bitrot.

shankari: Note also that we don't have to support a `.env` with a single tag. There are ways of editing files programmatically that don't involve `>>`

shankari (approved):

shankari:
Squash merging again, with 90 commits for 13 files.

PR #114, opened 2024-03-26
Map - show map only when there is trip data

jiji14:
### Overview
The current map page shows the map feature even though there is no trip data, which can appear as a bug to users. To prevent confusion, I added the 'no data' text when there is no trip data instead of displaying an empty map.

<br />

### screenshots
[ No trip data ]
Screenshot 2024-03-26 at 11 45 35 AM

Screenshot 2024-03-26 at 11 45 18 AM

### DIFF
--- a/pages/map.py
+++ b/pages/map.py
@@ -222,7 +222,9 @@ layout = html.Div(
                 dcc.Dropdown(id='user-mode-dropdown', multi=True),
             ], style={'display': 'block'})
         ]),
-
+        dbc.Row(
+            html.H4("No trip data available for the selected date", id="no-trip-text")
+        ),
         dbc.Row(
             dcc.Graph(id="trip-map")
         ),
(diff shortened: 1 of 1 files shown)


jiji14:
@JGreenlee Please review this when you have time :)

pages/map.py
```diff
@@ -324,13 +326,19 @@ def process_trips_group(trips_group):
 
 @callback(
     Output('store-trips-map', 'data'),
+    Output('no-trip-text', 'style'),
+    Output('trip-map', 'style'),
     Input('store-trips', 'data'),
 )
 def store_trips_map_data(trips_data):
+    # if there is no trips_data for selected date, show 'no-trip-text'
+    if trips_data == None or 'length' not in trips_data or trips_data['length'] == 0:
```
JGreenlee, line 335: This can be simplified:
```suggestion
    if not trips_data:
```

Empty lists and dictionaries are falsy in Python, so whether it is `{}` `[]` or `None`, this expression treats it the same.

By the way if you did need to specifically check the length, the usual way to do it is `len(trips_data)`

JGreenlee (commented):
This works well and helps UX!

There are many different ways you could conditionally render things in Dash. Although setting `display='none'` works, I think it might be more elegant to return the "No trip data available..." text from one of the functions.
I'm providing you with a patch file showing how I would I have done it. But it is up to you which method seems cleaner or more elegant

[conditional-map-alternative.patch](https://github.com/e-mission/op-admin-dashboard/files/14779745/conditional-map-alternative.patch)

JGreenlee (approved):

PR #117, opened 2024-06-14
expand filters to be able to exclude any subgroup(s)

JGreenlee:
We had a single checkbox to exclude test users, which worked fine for that purpose. But for programs with multiple subgroups we want to be able to filter by any of the subgroups

I modified this into a multi dropdown which has all subgroups from the config (or just 'test' if the config doesn't define subgroups)

I also made the default value reflect metrics.include_test_users from the config.


JGreenlee:
Here is what the multi dropdown looks like

image

<br>

image

<br>

image

PR #119, opened 2024-08-23
Update certs

MukuFlash03:
Updating certificates to use latest ones to match internal images


shankari:
@MukuFlash03 I don't think this is necessary, since we have already incorporated the certs into the server image (https://github.com/e-mission/e-mission-server/pull/976) and the dashboard image is based on the server image. Please let me know if this is incorrect.

PR #121, opened 2024-09-05
Fix Callback Error Updating store-trips.data - TypeError: 'float' object is not subscriptable fixed 

TeachMeTW:
### Description

We are encountering a `TypeError` while updating `store-trips.data`. The error traceback indicates that there is an issue with the function applied to the `ble_sensed_summary` column in the DataFrame. The error occurs because a `float` object is being treated as if it were a dictionary.

### Traceback
```
Callback error updating store-trips.data Traceback (most recent call last): File "/usr/src/app/app_sidebar_collapsible.py", line 318, in update_store_trips df, user_input_cols = query_confirmed_trips(start_date, end_date, timezone) File "/usr/src/app/utils/db_utils.py", line 101, in query_confirmed_trips df["data.primary_ble_sensed_mode"] = df.ble_sensed_summary.apply(get_max_mode_from_summary) File "/root/miniconda-23.5.2/envs/emission/lib/python3.9/site-packages/pandas/core/series.py", line 4771, in apply return SeriesApply(self, func, convert_dtype, args, kwargs).apply() File "/root/miniconda-23.5.2/envs/emission/lib/python3.9/site-packages/pandas/core/apply.py", line 1123, in apply return self.apply_standard() File "/root/miniconda-23.5.2/envs/emission/lib/python3.9/site-packages/pandas/core/apply.py", line 1174, in apply_standard mapped = lib.map_infer( File "pandas/_libs/lib.pyx", line 2924, in pandas._libs.lib.map_infer File "/usr/src/app/utils/db_utils.py", line 97, in <lambda> get_max_mode_from_summary = lambda md: max(md["distance"], key=md["distance"].get) if len(md["distance"]) > 0 else "INVALID" TypeError: 'float' object is not subscriptable
```

### Issue

The `TypeError` occurs because the lambda function `get_max_mode_from_summary` expects `md` to be a dictionary with a `"distance"` key that maps to another dictionary. However, it appears that some entries in `ble_sensed_summary` are `float` values, which are not subscriptable and thus lead to this error.

### Steps to Reproduce

1. Call the `update_store_trips` function with appropriate parameters.
2. Observe the error traceback indicating the `TypeError`.

### Expected Behavior

The `ble_sensed_summary` column should be processed correctly, and the `get_max_mode_from_summary` function should handle various data types properly without raising an error.

### Suggested Fix

Update the `get_max_mode_from_summary` lambda function to handle cases where `md` is not a dictionary. For example:

```python
get_max_mode_from_summary = lambda md: (
    "INVALID" if not isinstance(md, dict) or "distance" not in md or not isinstance(md["distance"], dict) else
    (max(md["distance"], key=md["distance"].get) if len(md["distance"]) > 0 else "INVALID")
)

Consider adding data validation or cleaning steps before applying the lambda function.

### CODE REVIEW

**Comment by TeachMeTW** [comment]
Utilized Black Formatter (for formatting purposes)
Modified mongodump

**Comment by TeachMeTW** [comment]
Use Case of Updated Script
```./load_mongodump.sh ~/Downloads/openpath_prod_ca_ebike_may_29.tar.gz Stage_database op-admin-dashboard-db-1 openpath_prod_ca_ebike ```

**Review by JGreenlee — COMMENTED** [review]
The error handling is great.

As you have seen, this project doesn't use Black or adhere to PEP8. Although that's not ideal, could you revert the formatting changes?
Standardized code format is great but when formatting changes are done at the same time as other changes, it is harder to review

If we were to enforce a code style, we would need to decide on it as a team, then update the whole codebase at once, independently from other changes
(we did this on e-mission-phone using Prettier, the JS equivalent of Black)

When I make changes to python code, I do try to adhere to PEP8 on the lines I change. But I don't worry about the rest of the file

File: docker/load_mongodump.sh
Comment (JGreenlee) [inline docker/load_mongodump.sh:1]: This is certainly a more elegant solution than the use case I showed you yesterday.

I do want to check if there's a cleaner/better way though. I think there's an environment variable that can be set to control what database is used by the dashboard (instead of the default `Stage_database`)

Comment (TeachMeTW) [reply docker/load_mongodump.sh:1]: @JGreenlee I looked everywhere for the Stage_database var but can't seem to find it; the compose file doesn't seem to point towards a certain DB with the exception of `study config`. I also reverted the formatting.

Comment (TeachMeTW) [reply docker/load_mongodump.sh:1]: I also modified the script due to another reason -- after loading the massive dump, I wanted to load in the smaller dataset; however, it kept the old data/did not drop/clean before loading in new data. Maybe there ought to be 2 loads? One that overwrites and one that appends? What do you think.

Comment (TeachMeTW) [reply docker/load_mongodump.sh:1]: @JGreenlee the ./load_mongodump.sh script was modified to allow user to select DB to be used and cleans it beforehand.

Use case:
`./load_mongodump.sh   <dump_file> <database_name> <docker_container_name> <collection_name>`
`./load_mongodump.sh ~/Downloads/openpath_prod_ca_ebike_may_29.tar.gz testing op-admin-dashboard-db-1 openpath_prod_ca_ebike`

is this what you meant?

Comment (JGreenlee) [reply docker/load_mongodump.sh:1]: The original script uses the same database name as the dump has (ie the directory name once the dump tar.gz is extracted). That is what we want. We don't need to add extra arguments.
The feature that is nice about your version is that it clears out existing db before attempting to restore. If you can do that without adding extra arguments go for it. 
Otherwise the script is fine how it was before.

Comment (TeachMeTW) [reply docker/load_mongodump.sh:1]: @JGreenlee Simplified the script; 

Use case:
`./load_mongodump.sh   <dump_file>`

Reads the DB from the Yaml file ensuring it plops it in the right place and clears it out before restoring. Removed excess args

Comment (shankari) [inline docker/load_mongodump.sh:1]: Nice and elegant! However, I have a suggestion to make it even more elegant. With the current approach, the user still needs to edit the `docker-compose` to put in the database name (e.g. `openpath_prod_ca_ebike`). I think it would be even better if they didn't have to do that.

So I would suggest flipping the flow.
- Have the `docker-compose` have a placeholder (e.g. mongodb://db:27017/DB_NAME)
- Determine the database name automatically from the mongodump
- Fill in the DB_NAME placeholder with the actual DB name

How can you determine the database name automatically from the mongodump? Something like `tar tzvf <mongodump> | head -n 1` should work since the directory name in the mongodump is the DB name.

Thoughts?

File: docker-compose-dev.yml [hunk 1]
```diff
@@ -14,7 +13,7 @@ services:
       DASH_DEBUG_MODE: "True"
       DASH_SILENCE_ROUTES_LOGGING: "False"
       DASH_SERVER_PORT: 8050
-      DB_HOST: db
+      DB_HOST: mongodb://db:27017/openpath_prod_ca_ebike
```
shankari, line 16: you can remove 27017 since that is the default port

shankari (commented):
My comments are only about the `load_mongodump` part

docker/load_mongodump.sh
TeachMeTW: @shankari I have implemented the change; same use case as `./load_mongodump.sh   <dump_file>`, but now it modifies/overwrites the compose DB_HOST based on the DB_NAME from the tar.

shankari:
@TeachMeTW can you pull out the loadmongodump changes into a separate PR (e.g. use `git checkout` to an earlier point and make a new branch)? I am happy to merge that now.

For the `ble_summary`, I would like to make sure that your approach is consistent with Jack and Abby's, so that can come in a follow-on PR.

docker-compose-dev.yml.bak
shankari, line 1: Also, remove this file (unnecessary)

docker-compose-dev.yml
```diff
@@ -30,7 +29,7 @@ services:
       - ./app_sidebar_collapsible.py:/usr/src/app/app_sidebar_collapsible.py
     deploy:
       restart_policy:
-          condition: on-failure
+        condition: on-failure
```
shankari, line 32: extraneous change, please revert to avoid commit churn

docker/load_mongodump.sh
```diff
@@ -1,10 +1,72 @@
...
+# Extract database name from the mongodump file
+TEMP_DIR=$(mktemp -d)
+tar -xf "$MONGODUMP_FILE" -C "$TEMP_DIR"
+DB_NAME=$(find "$TEMP_DIR/dump" -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)
+
+if [ -z "$DB_NAME" ]; then
+    echo "Error: Failed to extract database name from mongodump."
+    exit 1
+fi
+
```
shankari, line 48: This works, but something `tar tzvf ... | head -n 1` would be better, since we won't need to extract the fully mongodump locally - the extra space and I/O may matter for large files.

Can you clarify why 

```
DB_NAME=`tar tzvf ... | head -n 1`
```

doesn't work?

shankari (changes requested):

docker/load_mongodump.sh
TeachMeTW: @shankari For some reason when I did the tar tzvf, it couldn't find the name. I tried with grew, awk, sed, etc but still came out empty handed despite it clearly showing 'open-ca-ebike'. I could try again. Moved this to #122

shankari:
Since I merged #122 to master, this only has the actual functionality changes. @JGreenlee and @Abby-Wheelis, I will let you review this before it comes to me.

utils/db_utils.py
```diff
@@ -94,7 +94,26 @@ def query_confirmed_trips(start_date: str, end_date: str, tz: str):
         # Add primary modes from the sensed, inferred and ble summaries. Note that we do this
         # **before** filtering the `all_trip_columns` because the
         # *_section_summary columns are not currently valid
-        get_max_mode_from_summary = lambda md: max(md["distance"], key=md["distance"].get) if len(md["distance"]) > 0 else "INVALID"
```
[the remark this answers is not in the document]
TeachMeTW: @Abby-Wheelis I am not entirely sure what `INVALID` would represent here as opposed to unknown. I believe @shankari was the one who originally wrote this line of code that I refactored. I can change it to `UNKNOWN` to match #141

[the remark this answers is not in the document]
shankari: @TeachMeTW can you look at the commit where I wrote that code and see if I added in an explanation of why my choice? I have some vague recollection of the decision making, but the comments I wrote at the time are likely to be more clear.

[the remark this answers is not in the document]
TeachMeTW: @shankari @Abby-Wheelis based on existing comments, this is how I interpreted it.

`INVALID`: used to signify that the data or result is not just unknown but invalid. It implies that something went wrong or the data provided does not meet the expected format or criteria. It's a more explicit signal that the data is unusable.

`UNKNOWN`:  indicates that the mode is not known, but does not necessarily imply that there was an error or problem with the data. It often means that the information is not available or could not be determined but is expected to be a valid state to encounter.

Should we keep both Invalid and unknown or just use one?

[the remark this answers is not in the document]
Abby-Wheelis: I think we can keep both, thank you for the explanation! Hopefully once we figure out [#1088](https://github.com/e-mission/e-mission-docs/issues/1088) and resolve the underlying data issues there will be less `INVALID` sense modes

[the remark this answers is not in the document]
TeachMeTW: @shankari if possible please resolve the issue if no further reviews/changes are needed

JGreenlee (approved):
I agree with the reasoning of using `INVALID` rather than `UNKNOWN`

`UNKNOWN` exists in the data for trips where a mode could not be determined from the sensed information

`INVALID` is used to mark trips that are missing the ble summary. `INVALID` is not stored in the data model

shankari (approved):
LGTM!

nit/future fix: If we are creating a multi-line function anyway, I don't think it needs to be a lambda, we can make it a real function which may be easier to read.

But that is a style preference that I don't feel that strongly about so merging this anyway

shankari:
@JGreenlee @TeachMeTW I think this will address some of the issues with loading data on older deployments (e.g. ccebike/smart-commute/mm-masscec etc). We should spot check that once I have deployed to production.

PR #122, opened 2024-09-07
Enhance MongoDB Dump Restoration Script with Error Handling and Configuration Updates

TeachMeTW:
#### Description:
This pull request refactors the existing MongoDB dump restoration script to include additional features and improvements. The updated script now provides better error handling, debugging information, and integrates with a Docker Compose configuration file to dynamically update the database host configuration.

#### Changes:
1. **Script Header**:
   - Added `#!/bin/bash` shebang to specify the script interpreter.

2. **Directory and Configuration File**:
   - Determined the script directory and configuration file path dynamically.
   - Configured the path to `docker-compose-dev.yml` one level up from the script's directory.

3. **Argument Validation**:
   - Added checks to ensure exactly one argument (the path to the MongoDB dump file) is provided.
   - Added usage instructions and error handling if the argument is missing.

4. **File Existence Checks**:
   - Verified the existence of the provided MongoDB dump file.
   - Verified the existence of the Docker Compose configuration file.

5. **Configuration File Details**:
   - Outputted details about the Docker Compose configuration file for debugging.

6. **Database Name Extraction**:
   - Extracted the database name from the MongoDB dump file to update the Docker Compose configuration file accordingly.

7. **Configuration File Update**:
   - Updated the `DB_HOST` in the Docker Compose configuration file with the actual database name.

8. **Docker Operations**:
   - Printed debug information about the Docker operations.
   - Copied the MongoDB dump file to the Docker container.
   - Cleared existing databases in the Docker container, except for `admin` and `local`.
   - Restored the MongoDB dump to the appropriate database.

9. **Temporary Directory Cleanup**:
   - Removed the temporary directory created for extracting the dump file.


docker/load_mongodump.sh
```diff
@@ -1,10 +1,70 @@
...
+cat "$CONFIG_FILE"
+
+echo "Copying file to Docker container"
+docker cp "$MONGODUMP_FILE" op-admin-dashboard-db-1:/tmp
+
+FILE_NAME=$(basename "$MONGODUMP_FILE")
 
-FILE_NAME=`basename $MONGODUMP_FILE`
+echo "Clearing existing database"
+docker exec op-admin-dashboard-db-1 bash -c 'mongo --eval "db.getMongo().getDBNames().forEach(function(d) { if (d !== \"admin\" && d !== \"local\") db.getSiblingDB(d).dropDatabase(); })"'
```
shankari, line 65: we now already have DB_NAME, so I don't think you need to use `getDBNames`, right?
Can't you just do something like:

```
mongo $DB_NAME --eval "db.dropDatabase()"
```

from https://stackoverflow.com/a/8857373

shankari (changes requested):
Almost there, just want to simplify it a bit.

docker/load_mongodump.sh
TeachMeTW: @shankari You're correct, I seemed to have overlooked that. It should be good now?

shankari (approved):
LGTM!

shankari:
🥳 Congratulations on the first merge! 🎉

shankari:
@iantei can you make a similar change to the public dashboard as well?

iantei:
Sure, I will make similar changes to the public dashboard.

docker-compose-dev.yml
```diff
@@ -14,7 +14,7 @@ services:
       DASH_DEBUG_MODE: "True"
       DASH_SILENCE_ROUTES_LOGGING: "False"
       DASH_SERVER_PORT: 8050
-      DB_HOST: db
+      DB_HOST: mongodb://db/openpath_prod_ca_ebike
```
JGreenlee, line 17: Was this change intentional? Shouldn't the default still be `Stage_database` ? @shankari @TeachMeTW

JGreenlee: Ah, I see this gets overwritten in the new version of the script so this value is arbitrary.

I still think the placeholder should be something other than `openpath_prod_ca_ebike` to prevent confusion

shankari: Ah good point. I had suggested that it be a placeholder like `REPLACE_ME_DB`, but didn't validate that it actually changed to be that. Agree that this should be changed to a placeholder in cleanup.

TeachMeTW: @shankari @JGreenlee It was changed to a placeholder but due to running the script, it changed once more (basically everytime you run the script the db location changes)

We could fix this in 2 ways I believe:

1. .gitignore the yaml from now on
2. Modify the script to revert the modification after launching compose?

shankari: I would suggest `.gitignore` and add a section to the `README` highlighting that if anybody does change it, they should use `git add -f <filename>`

That optimizes for the common case (people using the docker-compose) and drops the burden of doing extra work on the uncommon case (people who are modifying it) and who are also more likely to understand and notice it.

TeachMeTW: @shankari would you prefer a new pull request or directly change the repo since this PR was already merged?

shankari: @TeachMeTW you can't directly change the repo; you don't have write access to it. It has to be a PR 😄

TeachMeTW: @shankari moved to #123

shankari:
@TeachMeTW the changes in this PR don't seem to have "stuck" to master.
Looking at the commit log, I think that they may have inadvertently gotten overwritten as part of some rebasing
the only commit I see to that file is one from Aug 2023
https://github.com/e-mission/op-admin-dashboard/commits/master/docker/load_mongodump.sh

Can you please restore the changes?

TeachMeTW:
> @TeachMeTW the changes in this PR don't seem to have "stuck" to master. Looking at the commit log, I think that they may have inadvertently gotten overwritten as part of some rebasing to clean up the commit history.
> 
> the only commit I see to that file is one from Aug 2023 https://github.com/e-mission/op-admin-dashboard/commits/master/docker/load_mongodump.sh
> 
> Can you please restore the changes?

@shankari addressed in https://github.com/e-mission/op-admin-dashboard/pull/158

PR #123, opened 2024-09-09
Update `.gitignore` and Add README Instructions

TeachMeTW:
### Overview

This pull request includes updates to the `.gitignore` file to ignore `docker-compose.yml` and adds instructions to the `README.md` for handling changes to this file. These changes are aimed at optimizing the workflow for most users and providing guidance for those who need to make exceptions.

### Changes Made

1. **Updated `.gitignore`**
   - Added `docker-compose-dev.yml` to the `.gitignore` file to prevent it from being tracked by Git. This ensures that local configuration changes do not interfere with the project’s standard setup.

2. **Enhanced `README.md`**
   - Added a section to the `README.md` to explain the reasoning behind ignoring `docker-compose-dev.yml`.
   - Included instructions for developers on how to commit changes to `docker-compose-dev.yml` if necessary, using the `-f` (force) option with `git add`.

3. **Implemented Placeholder for Compose**

### DIFF
--- a/docker/load_mongodump.sh
+++ b/docker/load_mongodump.sh
@@ -10,6 +10,7 @@ CONFIG_FILE="$SCRIPT_DIR/../docker-compose-dev.yml"
 if [ "$#" -ne 1 ]; then
     echo "Usage: $0 <mongodump-file>"
     echo "  <mongodump-file> : The path to the MongoDB dump file to be restored."
+    echo " run git add -f <docker compose file> after using this command"
     exit 1
 fi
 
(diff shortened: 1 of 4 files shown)


docker/load_mongodump.sh
```diff
@@ -10,6 +10,7 @@ CONFIG_FILE="$SCRIPT_DIR/../docker-compose-dev.yml"
 if [ "$#" -ne 1 ]; then
     echo "Usage: $0 <mongodump-file>"
     echo "  <mongodump-file> : The path to the MongoDB dump file to be restored."
+    echo " run git add -f <docker compose file> after using this command"
```
shankari, line 13: @TeachMeTW the rest of this change is fine, but this seems weird. Why are we putting in instructions for the **uncommon** case into the load script? If people change the `docker-compose.dev.yml` because of the load script, they shouldn't be checking it in.

They should only be checking it in if they went in and edited it directly.

shankari (changes requested):

docker/load_mongodump.sh
TeachMeTW: @shankari Fixed it, good catch, I think I autopiloted that line

shankari (approved):

shankari:
Squash merging to avoid commit churn.

PR #124, opened 2024-09-09
Performance Improvements in User Stats Processing

TeachMeTW:
## Description

This pull request introduces several performance improvements to the data page:

1. **Timer Integration**: Added timing to measure the performance of the data page.
2. **Lazy Loading**: Added processing of UUIDs in batches of 10.

### 1. Timer Integration

Added timing to measure the execution time of the `add_user_stats` function and other functions.

### 2. Lazy Loading Implementation

Replaced sequential user processing with batch processing; batches can be set in add_user_stats in db_utils for default value changes or in render_content in data.py

Each batch of 10 takes about ~7 seconds of processing.


TeachMeTW:
This PR would need a squash due to my extensive commits

shankari:
@JGreenlee @TeachMeTW I looked at this a while ago and came up with a plan (or two) but didn't have time to implement. The key here is to really dig into the user model and understand that the user stats consists of two separate parts:
- the user profile: which is stored as a single entry
- the user stats: which are computed from the trip table (e.g. % labeled, last trip, first trip, etc)

Reading the first is super fast, because it is a single small table
Reading the second is very slow because we have to essentially query for the entire trip table

We should fix this by:
- lazy loading the user stats part: I looked up plotly and it is possible to update data as new fields are computed (e.g. something like https://community.plotly.com/t/extend-or-append-data-instead-of-update/8898/4). I might have filed an issue for this as well but can't find it now.
- pre-computing the user stats and also storing them in the user profile: ideally we would do this as part of the pipeline OR a separate script that runs once a day. Note that we will also need a backwards compat script if we choose this.

shankari:
Ah I think it was `Patch` that I originally found for lazily updating data
https://plotly.com/blog/partial-properties/

TeachMeTW:
@shankari I made progress in regards to the lazy loading, the only thing I'd need to figure out is how to stop it from refreshing the ui. I tried to use Patch() but seems to not work fully as intended yet; I'll keep at it.

See below:


https://github.com/user-attachments/assets/f904005c-8f83-4cbf-81a1-cd48d59b21e5

TeachMeTW:
@shankari Progress, discussed with @JGreenlee to help resolve some of the issues I was facing; now loads without reloading the entire thing. 

https://github.com/user-attachments/assets/bd12efa9-4dde-442f-b248-ef8a04ad4816

requirements.txt
shankari, line 1: extraneous change

pages/data.py
```diff
@@ -60,66 +67,110 @@ def update_store_trajectories(start_date: str, end_date: str, tz: str, excluded_
...
+
+        if new_data:
+            # Process and append the new data to the loaded store
+            processed_data = db_utils.add_user_stats(new_data, initial_batch_size)
+            loaded_data.extend(processed_data)
+
+            # Create a Patch object to append data progressively
+            patched_data = Patch()
+            patched_data['data'] = processed_data
+
```
shankari, line 112: I don't understand how this is used. I see that the `patched_data` object is created here, but I don't see it used anywhere else in this PR. I even see that line 129 references this `Patch` object in a comment, but I don't see any of the `Patch` object methods, such as `append`.

Are we actually using `patch`? If not, what are we doing for lazy loading?

utils/db_utils.py
```diff
@@ -94,7 +114,26 @@ def query_confirmed_trips(start_date: str, end_date: str, tz: str):
         # Add primary modes from the sensed, inferred and ble summaries. Note that we do this
         # **before** filtering the `all_trip_columns` because the
         # *_section_summary columns are not currently valid
-        get_max_mode_from_summary = lambda md: max(md["distance"], key=md["distance"].get) if len(md["distance"]) > 0 else "INVALID"
+        
+        # Check if 'md' is not a dictionary or does not contain the key 'distance'
```
shankari, line 118: @TeachMeTW shouldn't this have already been included in https://github.com/e-mission/op-admin-dashboard/pull/121

shankari (approved):
This seems to work, and I want to get it out to staging ASAP so that we can test and move to production (and unblock everybody). We can always return and polish later.

But I do have some questions for me to understand your changes better

shankari:
Squash merging this as well, make sure to pull the changes before starting on the next PR, @TeachMeTW

app_sidebar_collapsible.py
```diff
@@ -210,17 +210,19 @@ def make_controls():
...
 
-page_content = dcc.Loading(
-    type='default',
-    fullscreen=True,
-    children=html.Div(dash.page_container, style={
+# Dcc Loading removed for Data Page Lazy Loading.
+# TODO Figure out how to enable Loading on everything BUT Data Page UUIDs Tab
+page_content = html.Div(
+    dash.page_container, 
+    style={
```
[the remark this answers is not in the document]
TeachMeTW: Without this change the whole UI is refreshed until whole dataset is loaded, stopping users from doing anything ie video 1

utils/db_utils.py
TeachMeTW: That's what I thought so too so I'm not sure why it was not transferred. I synced and pulled but still showed this.

pages/data.py
TeachMeTW: I believe this was a relic of a prior iteration I made; I can remove/clean this section in another PR.

PR #125, opened 2024-09-10
Updating Readme - Using environment variables for AWS Cognito

MukuFlash03:
Updating docs to reference changes from the Redesign PR #113 

These mostly involve removal of config-fake.py and config.py file that contained the CognitoConfig class for setting AWS Cognito credentials.

We now set the values directly via Environment variables.


shankari (approved):
🥳   Approval on the first pass 🎉

PR #131, opened 2024-09-20
Pyg walker mod

TeachMeTW:


TeachMeTW:
@JGreenlee Per our prior talks, I visited https://docs.kanaries.net/pygwalker in order to see if we can disable `askviz` and other unnecessary features, but I cannot seem to find a solution. I believe we'd have to choose between older version (<3.9) or continue with the new version but put a disclaimer? Another possible solution is injected js/html before it is rendered to get rid of said unwanted features? Thoughts?

PR #133, opened 2024-09-24
UUE Temp Fix

TeachMeTW:
### 1. Admin Dashboard Toggle for Key List
A toggle has been added to the "Trajectories" tab in the admin dashboard. This will allow administrators to switch between viewing "Analysis/Recreated Location" data and raw "Background/Location" data. The toggle is initially hidden and only visible when the "Trajectories" tab is active.

### Purpose
This change is a high-priority temporary measure to restore functionality to UUE's admin dashboard. It enables them to access raw trajectory data, ensuring they remain unblocked while a more permanent solution is being worked on.

### Testing
- Verified that the toggle switch properly switches between the two data views.

PR #135, opened 2024-09-25
Refactored Loading UI for Improved User Experience

TeachMeTW:
## Refactored Loading UI for Improved User Experience

- Updated the loading UI to provide better visual feedback when switching between tabs or pages.
- Replaced the default Dash loading spinner with a semi-transparent version that allows users to see partial data while loading is in progress.
  
  - **Before**: The default opaque loading screen blocked content completely.
  - **Now**: Users can see the underlying data while the loading animation is active, providing a more seamless experience.

### Known Issue
- The lazy loading on the UUID page now triggers the loading animation, meaning users cannot fully interact with the data until it has completely loaded. This will be addressed in a future update, but is necessary for now.
- @shankari  On #133 , users would NOT know if data was loading between tabs/toggling the switch. There is no differentiating whether its in an error state or a loading state except for to look at the logs or see the browser title saying 'Updating'. This was due to the lack of loading animation (that I removed due to the lazy loading that I now reworked and replaced).

### DIFF
--- a/pages/data.py
+++ b/pages/data.py
@@ -29,6 +29,7 @@ layout = html.Div(
             dcc.Tab(label='Trajectories', value='tab-trajectories-datatable'),
         ]),
         html.Div(id='tabs-content'),
+        dcc.Store(id='selected-tab', data='tab-uuids-datatable'),  # Store to hold selected tab
         dcc.Interval(id='interval-load-more', interval=10000, n_intervals=0), # default loading at 10s, can be lowered or hightened based on perf (usual process local is 3s)
         dcc.Store(id='store-uuids', data=[]),  # Store to hold the original UUIDs data
         dcc.Store(id='store-loaded-uuids', data={'data': [], 'loaded': False})  # Store to track loaded data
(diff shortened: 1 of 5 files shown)


shankari:
I am going to squash merge these changes since they seem important, but in the future, I would like more of a description on the "why" of the changes. For example, why do we need to bump up plotly and dash? As an open-source project, part of being transparent is documenting the reason behind our changes.

Here's a recent example: https://github.com/e-mission/e-mission-server/commit/250c26ea3fe07ea5b0f5dea2ccf12418f395b824

pages/data.py
```diff
@@ -94,63 +101,73 @@ def update_store_trajectories(start_date: str, end_date: str, tz: str, excluded_
...
+                logging.debug(f"New batch loaded. Total loaded: {len(loaded_data)}")
 
         # Prepare the data to be displayed
         columns = perm_utils.get_uuids_columns()  # Get the relevant columns
         df = pd.DataFrame(loaded_data)
 
         if df.empty or not perm_utils.has_permission('data_uuids'):
             logging.debug("No data or permission issues.")
-            return html.Div([html.P("No data available or you don't have permission.")]), loaded_uuids_store, False
+            return html.Div([html.P("No data available or you don't have permission.")]), loaded_uuids_store, True
```
shankari, line 157: why did this need to be changed? what is the second return parameter?

shankari (approved):

shankari:
I actually pulled the logs and .gitignore seems to be the same. git won't let me commit it since nothing has changed.
Squash merging...

pages/data.py
TeachMeTW: @shankari, @JGreenlee improved the code by separating the callbacks for the trajectories switch from the main. The second return is just to store the existing uuids so that it doesn't have to be requeried; probably could go on a separate callback as well

TeachMeTW:
> I am going to squash merge these changes since they seem important, but in the future, I would like more of a description on the "why" of the changes. For example, why do we need to bump up plotly and dash? As an open-source project, part of being transparent is documenting the reason behind our changes.
> 
> Here's a recent example: [e-mission/e-mission-server@250c26e](https://github.com/e-mission/e-mission-server/commit/250c26ea3fe07ea5b0f5dea2ccf12418f395b824) in case you need inspiration 😄

@shankari Sounds good, I'll add it to future prs/issues. For this particular update, we needed to bump up Plotly and Dash because the new features in dcc.Loading help us address the existing issue effectively such as modifying its transparency and logic.

shankari:
This has been pushed to staging. Can you please test so that we can push to production next week?

PR #142, opened 2024-10-10
Fix: Retain Current Page in UUID Table During Lazy Loading

### LINKED ISSUE #129: UUID Table Does Not Retain Current Page During Lazy Loading in Data Tab
#### Description
With the new implementation of lazy loading on the data page, the UUID table does not retain the user's current page when updates occur. If a user is skimming through the table while data is still loading, the page resets to the first page with every dataset update. This behavior persists until the entire dataset is fully loaded, after which it behaves as expected (since it is no longer refreshed).

#### Problem
When lazy loading is in progress, users navigating the table are automatically returned to the first page with every update. This interrupts the user experience, particularly when trying to review or analyze data in real-time. This issue doesn't occur once the dataset has completely loaded, but while loading, it can be very disruptive and frustrating.

#### Impact
- **User Experience**: Annoying and inefficient, as users are forced back to the first page, disrupting their workflow.
- **Usability**: Makes it difficult to browse and explore partial data while the lazy loading is occurring.

#### Proposed Solution
Track the user's current page during the lazy loading process and maintain it throughout updates. The table should:
1. Retain the user's page position when the dataset updates.
2. Only refresh the data without resetting the view to page one.
3. Continue allowing navigation during loading without disruptions.

#### Tasks
- Investigate why the current page state is lost during data updates in lazy loading.
- Implement a solution to retain the user’s page position during lazy loading updates.
- Test across different use cases to ensure the page retention works seamlessly with the lazy loading process.

TeachMeTW:
This pull request fixes issue #129, where the UUID table did not retain the user's current page during lazy loading updates. With this fix, the page now stays on the user's current page even as new data is loaded.

**Problem**

Previously, when lazy loading was in progress, navigating through the UUID table would reset the user back to the first page upon each data update. This disrupted the user experience, especially when trying to review data in real-time before the entire dataset was loaded.

**Solution**

- Tracked the user's current page state during lazy loading.
- Modified the data update mechanism to retain the current page without resetting to page one.
- Ensured that navigation remains uninterrupted during data loading.

**Impact**

- **User Experience**: Improved, as users can now navigate the table without being forced back to the first page.
- **Usability**: Enhanced, allowing users to browse and explore data seamlessly during lazy loading.

**Testing**

- Verified that the current page is retained during data updates.
- Tested across various use cases to ensure consistent behavior during lazy loading.
- Confirmed that navigation works smoothly even when the dataset is still loading.

https://github.com/user-attachments/assets/715ccd84-98e7-424c-89d9-508e1146528d

**Closes:** Fixes #129

### DIFF
--- a/pages/data.py
+++ b/pages/data.py
@@ -126,12 +126,9 @@ def render_content(
 
     # Update selected tab
     selected_tab = tab
-    logging.debug(f"Callback - {selected_tab} Stage 1: Selected tab updated.")
     
     # Handle the UUIDs tab without fullscreen loading spinner
     if tab == 'tab-uuids-datatable':
-        start_time = time.time()
-        logging.debug(f"Callback - {selected_tab} Stage 2: Handling UUIDs tab.")
 
         # Ensure store_uuids contains the key 'data' which is a list of dictionaries
         if not isinstance(store_uuids, dict) or 'data' not in store_uuids:
@@ -152,9 +149,6 @@ def render_content(
         if not loaded_uuids_store.get('loaded', False):
             total_to_load = total_loaded + initial_batch_size
             total_to_load = min(total_to_load, len(uuids_list))  # Avoid loading more than available
-
-            logging.debug(f"Callback - {selected_tab} Stage 3: Loading next batch of UUIDs from {total_loaded} to {total_to_load}.")
-
             new_data = uuids_list[total_loaded:total_to_load]
 
             if new_data:
@@ -166,19 +160,15 @@ def render_content(
                 loaded_uuids_store['data'] = loaded_data  # Mark all data as loaded if done
                 loaded_uuids_store['loaded'] = len(loaded_data) >= len(uuids_list)
 
-                logging.debug(f"Callback - {selected_tab} Stage 4: New batch loaded. Total loaded: {len(loaded_data)}.")
-
         # Prepare the data to be displayed
         columns = perm_utils.get_uuids_columns()  # Get the relevant columns
         df = pd.DataFrame(loaded_data)
 
         if df.empty or not perm_utils.has_permission('data_uuids'):
-            logging.debug(f"Callback - {selected_tab} Error Stage: No data available or permission issues.")
             return html.Div([html.P("No data available or you don't have permission.")]), loaded_uuids_store, True
 
         df = df.drop(columns=[col for col in df.columns if col not in columns])
 
-        logging.debug(f"Callback - {selected_tab} Stage 5: Returning appended data to update the UI.")
         content = html.Div([
             populate_datatable(df, table_id='uuid-table', page_current=current_page),  # Pass current_page
             html.P(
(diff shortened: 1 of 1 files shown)


TeachMeTW:
Removed unnecessary logging

TeachMeTW:
@JGreenlee Please Review

pages/data.py
```diff
@@ -277,19 +282,20 @@ def update_dropdowns_trips(n_clicks, button_label):
...
+        id=table_id,
         # columns=[{"name": i, "id": i} for i in df.columns],
         data=df.to_dict('records'),
         export_format="csv",
         filter_options={"case": "sensitive"},
         # filter_action="native",
         sort_action="native",  # give user capability to sort columns
         sort_mode="single",  # sort across 'multi' or 'single' columns
-        page_current=0,  # page number that user is on
+        page_current=page_current,  # set to current page
```
JGreenlee, line 298: Note to self: this is basically the fix

pages/data.py
```diff
@@ -22,17 +22,22 @@
...
+        dcc.Tabs(
+            id="tabs-datatable",
+            value='tab-uuids-datatable',
+            children=[
+                dcc.Tab(label='UUIDs', value='tab-uuids-datatable'),
+                dcc.Tab(label='Trips', value='tab-trips-datatable'),
+                dcc.Tab(label='Demographics', value='tab-demographics-datatable'),
+                dcc.Tab(label='Trajectories', value='tab-trajectories-datatable'),
+            ]
+        ),
```
JGreenlee, line 34: formatting change

pages/data.py
```diff
@@ -98,69 +113,64 @@ def show_keylist_switch(tab):
...
     State('store-loaded-uuids', 'data'),  # Use State to track already loaded data
     State('store-loaded-uuids', 'loaded')  # Keep track if we have finished loading all data
 )
-def render_content(tab, store_uuids, store_excluded_uuids, store_trips, store_demographics, store_trajectories,
-                   start_date, end_date, timezone, n_intervals, key_list, loaded_uuids_store, all_data_loaded):
+def render_content(
+    tab, store_uuids, store_excluded_uuids, store_trips, store_demographics, store_trajectories,
+    start_date, end_date, timezone, n_intervals, key_list, current_page,
+    loaded_uuids_store, all_data_loaded
+):
```
JGreenlee, line 124: added `current_page`

JGreenlee (commented):
Looks like a fairly straightforward fix, along with some other logging and comment cleanups

pages/data.py
```diff
@@ -98,69 +113,64 @@ def show_keylist_switch(tab):
...
-
-            # Slice the list of UUIDs from the dict
             new_data = uuids_list[total_loaded:total_to_load]
 
             if new_data:
                 # Process and append the new data to the loaded store
                 processed_data = db_utils.add_user_stats(new_data, initial_batch_size)
                 loaded_data.extend(processed_data)
-
+
```
JGreenlee, line 158:
```suggestion

```
remove this whitespace that didn't exist before

pages/data.py
```diff
@@ -98,69 +113,64 @@ def show_keylist_switch(tab):
...
+):
     initial_batch_size = 10  # Define the batch size for loading UUIDs
 
     # Update selected tab
     selected_tab = tab
-    logging.debug(f"Selected tab: {selected_tab}")
+    
     # Handle the UUIDs tab without fullscreen loading spinner
     if tab == 'tab-uuids-datatable':
+
```
JGreenlee, line 132:
```suggestion
```

pages/data.py
```diff
@@ -22,17 +22,22 @@
...
+                dcc.Tab(label='UUIDs', value='tab-uuids-datatable'),
+                dcc.Tab(label='Trips', value='tab-trips-datatable'),
+                dcc.Tab(label='Demographics', value='tab-demographics-datatable'),
+                dcc.Tab(label='Trajectories', value='tab-trajectories-datatable'),
+            ]
+        ),
         html.Div(id='tabs-content'),
         dcc.Store(id='selected-tab', data='tab-uuids-datatable'),  # Store to hold selected tab
-        dcc.Interval(id='interval-load-more', interval=20000, n_intervals=0), # default loading at 10s, can be lowered or hightened based on perf (usual process local is 3s)
+        dcc.Interval(id='interval-load-more', interval=24000, n_intervals=0),  # Interval for loading more data
```
JGreenlee, line 37: Interval increased; ok for now until we implement a better solution that is not as fragile and does not depend on an arbitrary "magic number"

PR #149, opened 2024-10-21
Added Timings

TeachMeTW:
Add Timing Instrumentation to Functions

## Overview
Introduces timing instrumentation to functions across the codebase (home, data, db_utils). The aim is to improve performance tracking by measuring the execution time for different stages within each function. By leveraging `ect.Timer()`, we capture detailed timings and store them using `esdsq.store_dashboard_time()` for later analysis.

## Key Changes
- Timings have been added to functions handling data retrieval, processing, and rendering in different components.
- Each function now captures the time taken at crucial steps such as querying data, processing records, updating UI components, and generating plots.
- The total time for each function execution is also recorded, allowing better insights into performance bottlenecks.

## Purpose
These changes provide visibility into the time spent in each stage of function execution, allowing for performance optimization and debugging. The captured timing data can be used for monitoring the efficiency of data-heavy operations, ensuring that the system remains responsive and scalable.

https://github.com/user-attachments/assets/3e9c6309-8652-48a2-b917-2cfddbb6c810

PR #150, opened 2024-10-22
Added #121

TeachMeTW:


shankari:
@TeachMeTW can you highlight how this wasn't found in the testing for #149

shankari:
> One user (me) had a 504 timeout error while @shankari discovered it was due to this issue.

@TeachMeTW I would like to understand why you did not encounter this error **in local testing**

TeachMeTW:
> > One user (me) had an error while @shankari discovered it was due to this issue.
> > 
> > 
> > 
> > @TeachMeTW I would like to understand why you did not encounter this error **in local testing**

In my local testing, I'm not entirely sure why the error didn't occur. Here's what I did:

- **Function Invocation**: I invoked the relevant function that might be related to the issue.
- **Datasets Used**: I loaded two different datasets: `ca_e-bike` and `open_access`.
- **Testing Steps**:
  - Navigated to the **Trips** tab and changed the date.
  - Changed the date both in the **Data** page and the **Home** page separately to ensure comprehensive testing.

Throughout these steps, no errors were encountered. It's possible that the issue is specific to the production environment or certain data conditions that aren't present in my local setup.

**Next Steps to Investigate:**

1. **Data Discrepancies**:
   - **Stage vs. Local Data**: Verify if the stage datasets have any differences or edge cases not present locally.
   - **Data Integrity**: Check for any corrupted or unexpected data formats in the stage datasets.

2. **Reproduce with Staging Data**:
   - If possible, use a subset of the stage data in my local environment to see if the error can be replicated.

I'm open to any additional suggestions or insights that could help identify why the error isn't manifesting in my local tests but is affecting others. I'll make sure to ask @JGreenlee as to why I cannot reproduce the error

TeachMeTW:
Maybe it's because I haven't reset my data? I've been using the same loaded open access and ca e-bike since the very beginning? They may have been modified -- I can try re intaking them and test that

JGreenlee:
We had previously talked about this bug in https://github.com/e-mission/op-admin-dashboard/pull/121 and https://github.com/e-mission/em-public-dashboard/pull/141.
It is due to missing values in some datasets. It only affects certain programs.

My guess is that it affects Stage but it does not affect `ca_e-bike` or `open_access`. Thus, the local testing performed was not adequate to reproduce this issue.
I recommend performing local testing with a dump of Stage

JGreenlee (approved):
The commit in this PR appears to have been generated by copy-pasting code from #121. It would have been preferable to use `git cherry-pick`. However, I see that the indentation level has changed with the addition of Timer blocks, so maybe you tried `git cherry-pick` and it did not work.

Regardless, I manually compared this PR to #121 and they are the same except for the indentation level. Approved.

TeachMeTW:
> The commit in this PR appears to have been generated by copy-pasting code from #121. It would have been preferable to use `git cherry-pick`. However, I see that the indentation level has changed with the addition of Timer blocks, so maybe you tried `git cherry-pick` and it did not work.
> 
> Regardless, I manually compared this PR to #121 and they are the same except for the indentation level. Approved.

Yes I did try to cherry pick but there were merge conflicts in regards to the whole function

TeachMeTW:
@JGreenlee  @shankari  I reproduced with `openpath-stage-snapshot-oct-6` and was able to reproduce the error:

image

Seems Jack was right:

> We had previously talked about this bug in #121 and [e-mission/em-public-dashboard#141](https://github.com/e-mission/em-public-dashboard/pull/141). It is due to missing values in some datasets. It only affects certain programs.
> 
> My guess is that it affects Stage but it does not affect `ca_e-bike` or `open_access`. Thus, the local testing performed was not adequate to reproduce this issue. I recommend performing local testing with a dump of Stage

It is dataset dependent

PR #151, opened 2024-10-27
Revert Data Page

TeachMeTW:
# [Revert Baseline Changes and Apply Selective Updates]

## Description

- **Reverted changes from baseline**
- **Cherry-picking Changes:**
  - Attempted to cherry-pick changes
  - Copied and pasted changes that couldn't be cherry-picked or had significant commit churn/hard to discern
- **Bug Fixes:**
  - Addressed various bugs identified during development
- **Re-added Timings:**
  - Restored timing functionalities

## Testing

Performed the following tests to ensure functionality and stability:

1. **Initial Load**
2. **Navigation from Sidebar:**
   - Clicked on "Data"
3. **Tabs Functionality:**
   - Clicked on "Trips" tab
   - Clicked on "Demographics" tab
   - Clicked on "Trajectories" tab
4. **Map Features:**
   - From sidebar, clicked on "Map"
   - Selected "Density Heatmap"
   - Selected "Unlabeled"

**Result:** All functionalities operated without errors.

https://github.com/user-attachments/assets/b08847d3-5d62-4882-ad94-b78cf81956ba

### DIFF
--- a/app_sidebar_collapsible.py
+++ b/app_sidebar_collapsible.py
@@ -210,15 +210,29 @@ def make_controls():
              'flex-direction': 'column'}
   )
 
+
 page_content = dcc.Loading(
+    id='global-loading',
     type='default',
     fullscreen=True,
+    overlay_style={"visibility": "visible", "filter": "blur(2px)"},
+    style={"background-color": "transparent"},
     children=html.Div(dash.page_container, style={
         "margin-left": "5rem",
         "margin-right": "2rem",
         "padding": "2rem 1rem",
     })
 )
+
+
+@app.callback(
+    Output('global-loading', 'display'),
+    Input('interval-load-more', 'disabled'),
+)
+def hide_spinner_while_loading_batch(interval_disabled):
+    if interval_disabled:
+        return 'auto'
+    return 'hide'
 
 
 def make_home_page(): return [
--- a/app_sidebar_collapsible.py
+++ b/app_sidebar_collapsible.py
@@ -210,12 +210,11 @@ def make_controls():
              'flex-direction': 'column'}
   )
 
-
 page_content = dcc.Loading(
     id='global-loading',
     type='default',
     fullscreen=True,
-    overlay_style={"visibility": "visible", "filter": "blur(2px)"},
+    overlay_style={"visibility": "visible", "filter": "opacity(0.5)"},
     style={"background-color": "transparent"},
     children=html.Div(dash.page_container, style={
         "margin-left": "5rem",
--- a/pages/data.py
+++ b/pages/data.py
@@ -172,157 +172,172 @@ def render_content(
     start_date, end_date, timezone, n_intervals, key_list, current_page,
     loaded_uuids_store, all_data_loaded
 ):
-    initial_batch_size = 10  # Define the batch size for loading UUIDs
+    with ect.Timer() as total_timer:
+        initial_batch_size = 10  # Define the batch size for loading UUIDs
 
-    # Update selected tab
-    selected_tab = tab
-    logging.debug(f"Callback - {selected_tab} Stage 1: Selected tab updated.")
-    
-    # Handle the UUIDs tab without fullscreen loading spinner
-    if tab == 'tab-uuids-datatable':
-        start_time = time.time()
-        logging.debug(f"Callback - {selected_tab} Stage 2: Handling UUIDs tab.")
+        # Stage 1: Update selected tab
+        selected_tab = tab
+        logging.debug(f"Callback - {selected_tab} Stage 1: Selected tab updated.")
 
-        # Ensure store_uuids contains the key 'data' which is a list of dictionaries
-        if not isinstance(store_uuids, dict) or 'data' not in store_uuids:
-            logging.error(f"Expected store_uuids to be a dict with a 'data' key, but got {type(store_uuids)}")
-            return html.Div([html.P("Data structure error.")]), loaded_uuids_store, True
+        # Handle the UUIDs tab without fullscreen loading spinner
+        if tab == 'tab-uuids-datatable':
+            with ect.Timer() as handle_uuids_timer:
+                logging.debug(f"Callback - {selected_tab} Stage 2: Handling UUIDs tab.")
 
-        uuids_list = store_uuids['data']
+                # Ensure store_uuids contains the key 'data' which is a list of dictionaries
+                if not isinstance(store_uuids, dict) or 'data' not in store_uuids:
+                    logging.error(f"Expected store_uuids to be a dict with a 'data' key, but got {type(store_uuids)}")
+                    return html.Div([html.P("Data structure error.")]), loaded_uuids_store, True
 
-        # Ensure uuids_list is a list for slicing
-        if not isinstance(uuids_list, list):
-            logging.error(f"Expected store_uuids['data'] to be a list but got {type(uuids_list)}")
-            return html.Div([html.P("Data structure error.")]), loaded_uuids_store, True
+                uuids_list = store_uuids['data']
 
-        loaded_data = loaded_uuids_store.get('data', [])
-        total_loaded = len(loaded_data)
+                # Ensure uuids_list is a list for slicing
+                if not isinstance(uuids_list, list):
+                    logging.error(f"Expected store_uuids['data'] to be a list but got {type(uuids_list)}")
+                    return html.Div([html.P("Data structure error.")]), loaded_uuids_store, True
 
-        # Handle lazy loading
-        if not loaded_uuids_store.get('loaded', False):
-            total_to_load = total_loaded + initial_batch_size
-            total_to_load = min(total_to_load, len(uuids_list))  # Avoid loading more than available
+                loaded_data = loaded_uuids_store.get('data', [])
+                total_loaded = len(loaded_data)
 
-            logging.debug(f"Callback - {selected_tab} Stage 3: Loading next batch of UUIDs from {total_loaded} to {total_to_load}.")
+                # Handle lazy loading
+                if not loaded_uuids_store.get('loaded', False):
+                    total_to_load = total_loaded + initial_batch_size
+                    total_to_load = min(total_to_load, len(uuids_list))  # Avoid loading more than available
 
-            new_data = uuids_list[total_loaded:total_to_load]
+                    logging.debug(f"Callback - {selected_tab} Stage 3: Loading next batch of UUIDs from {total_loaded} to {total_to_load}.")
 
-            if new_data:
-                # Process and append the new data to the loaded store
-                processed_data = db_utils.add_user_stats(new_data, initial_batch_size)
-                loaded_data.extend(processed_data)
-                
-                # Update the store with the new data
-                loaded_uuids_store['data'] = loaded_data  # Mark all data as loaded if done
-                loaded_uuids_store['loaded'] = len(loaded_data) >= len(uuids_list)
+                    new_data = uuids_list[total_loaded:total_to_load]
 
-                logging.debug(f"Callback - {selected_tab} Stage 4: New batch loaded. Total loaded: {len(loaded_data)}.")
+                    if new_data:
+                        # Process and append the new data to the loaded store
+                        processed_data = db_utils.add_user_stats(new_data, initial_batch_size)
+                        loaded_data.extend(processed_data)
+                        
+                        # Update the store with the new data
+                        loaded_uuids_store['data'] = loaded_data  # Mark all data as loaded if done
+                        loaded_uuids_store['loaded'] = len(loaded_data) >= len(uuids_list)
 
-        # Prepare the data to be displayed
-        columns = perm_utils.get_uuids_columns()  # Get the relevant columns
-        df = pd.DataFrame(loaded_data)
+                        logging.debug(f"Callback - {selected_tab} Stage 4: New batch loaded. Total loaded: {len(loaded_data)}.")
 
-        if df.empty or not perm_utils.has_permission('data_uuids'):
-            logging.debug(f"Callback - {selected_tab} Error Stage: No data available or permission issues.")
-            return html.Div([html.P("No data available or you don't have permission.")]), loaded_uuids_store, True
+                # Prepare the data to be displayed
+                columns = perm_utils.get_uuids_columns()  # Get the relevant columns
+                df = pd.DataFrame(loaded_data)
 
-        df = df.drop(columns=[col for col in df.columns if col not in columns])
+                if df.empty or not perm_utils.has_permission('data_uuids'):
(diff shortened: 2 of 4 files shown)


app_sidebar_collapsible.py
```diff
@@ -210,9 +210,13 @@ def make_controls():
              'flex-direction': 'column'}
   )
 
+
```
TeachMeTW, line 213: whitespace change

pages/data.py
```diff
@@ -111,123 +161,222 @@ def update_store_trajectories(start_date: str, end_date: str, tz: str, excluded_
...
+                data = store_trips.get("data", [])
                 columns = perm_utils.get_allowed_trip_columns()
-                columns.update(
-                    col['label'] for col in perm_utils.get_allowed_named_trip_columns()
-                )
-                columns.update(store_trips["userinputcols"])
+                columns.update(col['label'] for col in perm_utils.get_allowed_named_trip_columns())
+                columns.update(store_trips.get("userinputcols", []))
                 has_perm = perm_utils.has_permission('data_trips')
+
```
TeachMeTW, line 267: whitespace change

pages/data.py
```diff
@@ -111,123 +161,224 @@ def update_store_trajectories(start_date: str, end_date: str, tz: str, excluded_
...
-                if store_trajectories == {}:
-                    store_trajectories = update_store_trajectories(start_date, end_date, timezone, store_excluded_uuids)
-                data = store_trajectories["data"]
+            with ect.Timer() as handle_trajectories_timer:
+                (start_date_only, end_date_only) = iso_to_date_only(start_date, end_date)
+                # Fetch new data based on the selected key_list from the keylist-switch
+                if not store_trajectories or key_list:
+                    store_trajectories = update_store_trajectories(start_date_only, end_date_only, timezone, store_excluded_uuids, key_list)
+
+                data = store_trajectories.get("data", [])
```
TeachMeTW, line 340: Brought back key_list from UUE fix

utils/db_utils.py
```diff
@@ -424,100 +430,119 @@ def query_trajectories(start_date: str, end_date: str, tz: str):
...
-                with ect.Timer() as stage4_timer:
-                    time_format = 'YYYY-MM-DD HH:mm:ss'
-                    ts = esta.TimeSeries.get_time_series(user_uuid)
-                    start_ts = ts.get_first_value_for_field(
-                        key='analysis/confirmed_trip',
-                        field='data.end_ts',
-                        sort_order=pymongo.ASCENDING
+        # Stage 1: Define process_user
+        with ect.Timer() as stage1_timer:
+            def process_user(user):
```
shankari, line 438: why are we defining this within the timer? We want to measure execution time, not definition time.

utils/db_utils.py
```diff
@@ -424,100 +430,119 @@ def query_trajectories(start_date: str, end_date: str, tz: str):
...
-                        user['last_trip'] = arrow.get(end_ts).format(time_format)
+                return user
+        esdsq.store_dashboard_time(
+            "admin/db_utils/add_user_stats/define_process_user",
+            stage1_timer
+        )
+        
+        # Stage 2: Define batch_process
+        with ect.Timer() as stage2_timer:
+            def batch_process(users_batch):
```
shankari, line 505: Ditto

utils/db_utils.py
```diff
@@ -424,100 +430,119 @@ def query_trajectories(start_date: str, end_date: str, tz: str):
...
+                with ect.Timer() as batch_process_timer:
+                    with ThreadPoolExecutor() as executor:  # Adjust max_workers based on CPU cores
+                        futures = [executor.submit(process_user, user) for user in users_batch]
+                        processed_batch = [future.result() for future in as_completed(futures)]
                 esdsq.store_dashboard_time(
-                    "admin/db_utils/add_user_stats/get_last_trip_timestamp",
-                    stage5_timer
+                    "admin/db_utils/add_user_stats/batch_process",
+                    batch_process_timer
```
shankari, line 512: Again, renaming metrics will make it challenging to compare the before and the after.

shankari (changes requested):
Has this been approved by @JGreenlee? We need to deploy it to staging today so we can start the next week with the new round of data collection?

@TeachMeTW I still see a lot of whitespace changes and reformats as seen below. 
Why is this still problem? I thought you had removed your autoformatter.

Please revert and rebase your existing commits. If you don't, then either:
- I merge without squashing, so the commit history is messed up OR
- I squash merge, in which case, multiple changes are merged into one, which also messes up the commit history and makes it hard to revert individual changes

pages/data.py
```diff
@@ -56,18 +83,18 @@ def clean_location_data(df):
             )
 
     esdsq.store_dashboard_time(
-        "admin/db_utils/clean_location_data/total_time",
+        "admin/data/clean_location_data/total_time",
```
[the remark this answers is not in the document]
TeachMeTW: Alright, it seems I had a typo previously so this was the only spot that was misnamed as db_utils -- everything else was data. Your right it'd be hard to compare and analyze later but I believe we should address the typo at some point

pages/data.py
```diff
@@ -111,123 +161,224 @@ def update_store_trajectories(start_date: str, end_date: str, tz: str, excluded_
...
+    Input('uuids-page-current', 'data'),  # Current page number for UUIDs DataTable
+    State('store-loaded-uuids', 'data'),  # Use State to track already loaded data
+    State('store-loaded-uuids', 'loaded')  # Keep track if we have finished loading all data
 )
-def render_content(tab, store_uuids, store_excluded_uuids, store_trips, store_demographics, store_trajectories, start_date, end_date, timezone):
+def render_content(
+    tab, store_uuids, store_excluded_uuids, store_trips, store_demographics, store_trajectories,
+    start_date, end_date, timezone, n_intervals, key_list, current_page,
+    loaded_uuids_store, all_data_loaded
+):
```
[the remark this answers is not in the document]
TeachMeTW: There's an addition to new parameters like `key_list`, `n_intervals`, `current_page`

utils/db_utils.py
TeachMeTW: I named it differently because the functionality is different; what do you suggest -- keep the old name or just track total time?

utils/db_utils.py
TeachMeTW: Seems to be a dup/unnecessary as you suggest

TeachMeTW:
The whitespace/reformats were from the non-cherried reverts

TeachMeTW:
@shankari I rebased the whitespace change commits; first time doing this and great learning experience -- looks good on my tests so I don't think I broke anything.

shankari:
@TeachMeTW are the changes done? should I plan on re-reviewing?

TeachMeTW:
> @TeachMeTW are the changes done? should I plan on re-reviewing?

Yes please re review @shankari

pages/data.py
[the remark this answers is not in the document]
shankari: Some point != now; let's finish this report first.

pages/data.py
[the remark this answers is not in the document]
shankari: That's why I said 

> there are whitespace changes that are not related to the real change

there is no reason for the arguments to go to the next line

shankari:
@TeachMeTW you need to rebase a few more changes:
- https://github.com/e-mission/op-admin-dashboard/pull/151/commits/8f3302a5a76e9a4da6ea14e0a76d97255b9ab878 and https://github.com/e-mission/op-admin-dashboard/pull/151/commits/71a405717bd0edc7f15a11e90fc36f2caf6235df need to be squashed. There is really no reason to have merge conflicts in the commit history
- What is the full description of https://github.com/e-mission/op-admin-dashboard/pull/151/commits/7d20994ea1a7093ab5a6d2bbb751331d78f82e7a? Reverted to what? can it be squashed with previous changes?
- Similarly, what is the bug in https://github.com/e-mission/op-admin-dashboard/pull/151/commits/e4eed98574d15e3b84c4c06268172b08b535fec7 and how was it fixed?

Please also handle the remaining whitespace from above.

Please provide descriptions for your commit messages. `git commit -m` (single line commit message) is insufficient.

utils/db_utils.py
shankari: old name

TeachMeTW:
@shankari ready for re-review; addressed changes

pages/data.py
```diff
@@ -21,7 +21,7 @@
 intro = """## Data"""
 
 layout = html.Div(
-    [   
+    [
```
shankari, line 24: One more extraneous whitespace

pages/data.py
```diff
@@ -111,123 +156,222 @@ def update_store_trajectories(start_date: str, end_date: str, tz: str, excluded_
...
     Input('date-picker-timezone', 'value'),
+    Input('interval-load-more', 'n_intervals'),  # Interval to trigger the loading of more data
+    Input('keylist-switch', 'value'),  # Add keylist-switch to trigger data refresh on change
+    Input('uuids-page-current', 'data'),  # Current page number for UUIDs DataTable
+    State('store-loaded-uuids', 'data'),  # Use State to track already loaded data
+    State('store-loaded-uuids', 'loaded')  # Keep track if we have finished loading all data
 )
-def render_content(tab, store_uuids, store_excluded_uuids, store_trips, store_demographics, store_trajectories, start_date, end_date, timezone):
+def render_content(
+    tab, store_uuids, store_excluded_uuids, store_trips, store_demographics, store_trajectories, start_date, end_date, timezone, n_intervals, key_list, current_page, loaded_uuids_store, all_data_loaded
```
shankari, line 166: extraneous change moving the `tab, store_uuids...` to the next line. Please see 
https://github.com/e-mission/op-admin-dashboard/pull/151#discussion_r1821674799

Not sure how it was resolved when this still hasn't been changed to match.

shankari (changes requested):
Almost there! Don't forget to rebase after reverting the whitespace changes.

TeachMeTW:
@shankari Changes made

shankari:
@TeachMeTW 

> Don't forget to rebase after reverting the whitespace changes.

we want to keep a clean commit history

TeachMeTW:
@shankari Which commits would you suggesting rebasing?

shankari:
https://github.com/e-mission/op-admin-dashboard/pull/151/commits/ca19e4c47c849652c85ddede89fb17a97d7598d3 
We should not add the whitespace in a commit and then remove it in a separate commit unless there is a reason to do so.

JGreenlee (approved):
This looks clean to me now. Thanks @ shankari for reviewing yesterday while I was out sick.

@TeachMeTW I think it would be beneficial for us to look at your workflow and discuss methods that result in more intentional commits and fewer unwanted changes.

I get the sense that you have been taught to habitually commit after every small change. I generally think it's a good habit to commit frequently, but when you commit excessively you aren't able to effectively describe your changes in context and the history ends up being messier. With this approach, you will usually have to rebase or squash to end up with a clean history.

My recommended workflow is to commit only once an isolated task is completed. The idea is to only commit "viable" states of the codebase; i.e. don't commit something that is still "under construction".
To minimize unwanted changes, I highly recommend spot-checking changes **before** committing them. You can do this with `git diff` – although I prefer the Source Control features in VSCode because I find the interactive view more useful and you can easily isolate certain lines to commit or discard.

I'm happy to discuss this further or show you what I mean.

TeachMeTW:
@shankari So rebase ca19e4c into 8ac4bc?

TeachMeTW:
> This looks clean to me now. Thanks @ shankari for reviewing yesterday while I was out sick.
> 
> @TeachMeTW I think it would be beneficial for us to look at your workflow and discuss methods that result in more intentional commits and fewer unwanted changes.
> 
> I get the sense that you have been taught to habitually commit after every small change. I generally think it's a good habit to commit frequently, but when you commit excessively you aren't able to effectively describe your changes in context and the history ends up being messier. With this approach, you will usually have to rebase or squash to end up with a clean history.
> 
> My recommended workflow is to commit only once an isolated task is completed. The idea is to only commit "viable" states of the codebase; i.e. don't commit something that is still "under construction". To minimize unwanted changes, I highly recommend spot-checking changes **before** committing them. You can do this with `git diff` – although I prefer the Source Control features in VSCode because I find the interactive view more useful and you can easily isolate certain lines to commit or discard.
> 
> I'm happy to discuss this further or show you what I mean.

Got it, yeah I have been taught to frequently commit so it is a habit to break.

TeachMeTW:
@shankari The whitespace has been rebased

shankari:
> Got it, yeah I have been taught to frequently commit so it is a habit to break.

I think it is fine to commit frequently as long as you rebase at the end before you create the PR.
I also use `git diff` to "check" my local state before pushing.
For the record, when working with AFS for the EVerest project, they also rebased frequently 😄

shankari:
I don't think this is actually rebased correctly.

The whitespace was added in https://github.com/e-mission/op-admin-dashboard/pull/151/commits/d979a732fc95eb4dfa7dde40bc5320ac4a10ca05
and then removed in https://github.com/e-mission/op-admin-dashboard/pull/151/commits/5b84365f4740983ade8c2e730fbbd1393e134ca7

It's not like it doesn't exist at all 😄 

I am going to wait for a couple of hours to see if you can fix this but will plan on merging at around noon PT regardless since this is a fairly small issue.

@TeachMeTW please also add "testing done" for every set of changes.

shankari:
@TeachMeTW are you going to make any changes? I also still don't see any testing done.

TeachMeTW:
@shankari Not too sure how to fix this:

> I don't think this is actually rebased correctly.
> 
> The whitespace was added in [d979a73](https://github.com/e-mission/op-admin-dashboard/commit/d979a732fc95eb4dfa7dde40bc5320ac4a10ca05) and then removed in [5b84365](https://github.com/e-mission/op-admin-dashboard/commit/5b84365f4740983ade8c2e730fbbd1393e134ca7)
> 
> It's not like it doesn't exist at all 😄
> 
> I am going to wait for a couple of hours to see if you can fix this but will plan on merging at around noon PT regardless since this is a fairly small issue.
> 
> @TeachMeTW please also add "testing done" for every set of changes.

I have also been notified I am behind on my trainings so I have been working on that for the past few hours.

Furthermore, what do you mean by add 'Testing Done' -- via the commit or just here in the comments?
I've tested the latest commit and it works same as the video.

pages/data.py
```diff
@@ -21,7 +21,7 @@
 intro = """## Data"""
 
 layout = html.Div(
-    [
```
TeachMeTW, line 24: this whitespace change reappeared after I did the rebasing

TeachMeTW: thoughts @shankari
Basically my workflow was:
git rebase -i d241d37
edit said commit, removed the whitespace
git add 
git rebase continue
then git push

PR #153, opened 2024-11-08
 Integrate New User Statistics into Profile Retrieval

TeachMeTW:
## Overview

Enhances the user profile data retrieval process by incorporating newly stored statistics (`pipeline_range`, `total_trips`, `labeled_trips`, and `last_call`) into the existing `with` block. These statistics are now seamlessly integrated into the `user` dictionary, providing a more comprehensive set of user metrics for further analysis and dashboard visualization.

### 1. **Enhanced Profile Data Retrieval**

- **Existing Functionality**:
  - The `with` block previously retrieved basic profile information such as `platform`, `manufacturer`, `app_version`, `os_version`, and `phone_lang` from the database and assigned them to the `user` dictionary.

- **New Enhancements**:
  - **Pipeline Range**:
    - Retrieved `pipeline_range` data, including `start_ts` and `end_ts`, to understand the timeframe of the user's activity pipeline.
  - **Trip Counts**:
    - Extracted `total_trips` and `labeled_trips` to provide insights into the user's trip data and the extent of trip labeling.
  - **Last API Call Timestamp**:
    - Obtained `last_call` to track the most recent API interaction by the user.
  - **Optional Formatting**:
    - Included a formatted version of the `last_call` timestamp for better readability in reports and dashboards.

### 2. **Updated `with` Block Structure**

- **Integration of New Statistics**:
  - The `with` block now not only fetches and assigns basic profile attributes but also retrieves and assigns the newly stored statistics from the database.
  
  Awaiting https://github.com/e-mission/e-mission-server/pull/993

### DIFF
--- a/utils/db_utils.py
+++ b/utils/db_utils.py
@@ -437,57 +437,35 @@ def add_user_stats(user_data, batch_size=5):
         def process_user(user):
             with ect.Timer() as process_user_timer:
                 user_uuid = UUID(user['user_id'])
-                
-                # Fetch aggregated data for all users once and cache it
-                ts_aggregate = esta.TimeSeries.get_aggregate_time_series()
-    
-                # Fetch data for the user, cached for repeated queries
                 profile_data = edb.get_profile_db().find_one({'user_id': user_uuid})
+                # Fetch data for the user, cached for repeated queries
+                logging.info(f'keyspr: {profile_data}')
+                if not profile_data:
+                    profile_data = {}
+                # Assign existing profile attributes to the user dictionary
+                user['platform'] = profile_data.get('curr_platform')
+                user['manufacturer'] = profile_data.get('manufacturer')
+                user['app_version'] = profile_data.get('client_app_version')
+                user['os_version'] = profile_data.get('client_os_version')
+                user['phone_lang'] = profile_data.get('phone_lang')
                 
-                total_trips = ts_aggregate.find_entries_count(
-                    key_list=["analysis/confirmed_trip"],
-                    extra_query_list=[{'user_id': user_uuid}]
-                )
-                labeled_trips = ts_aggregate.find_entries_count(
-                    key_list=["analysis/confirmed_trip"],
-                    extra_query_list=[{'user_id': user_uuid}, {'data.user_input': {'$ne': {}}}]
-                )
+                # Assign newly stored statistics to the user dictionary
+                user['total_trips'] = profile_data.get('total_trips')
+                user['labeled_trips'] = profile_data.get('labeled_trips')
                 
-                user['total_trips'] = total_trips
-                user['labeled_trips'] = labeled_trips
-    
-                if profile_data:
-                    user['platform'] = profile_data.get('curr_platform')
-                    user['manufacturer'] = profile_data.get('manufacturer')
-                    user['app_version'] = profile_data.get('client_app_version')
-                    user['os_version'] = profile_data.get('client_os_version')
-                    user['phone_lang'] = profile_data.get('phone_lang')
-    
-                if total_trips > 0:
-                    ts = esta.TimeSeries.get_time_series(user_uuid)
-                    first_trip_ts = ts.get_first_value_for_field(
-                        key='analysis/confirmed_trip',
-                        field='data.end_ts',
-                        sort_order=pymongo.ASCENDING
-                    )
-                    if first_trip_ts != -1:
-                        user['first_trip'] = arrow.get(first_trip_ts).format(time_format)
-    
-                    last_trip_ts = ts.get_first_value_for_field(
-                        key='analysis/confirmed_trip',
-                        field='data.end_ts',
-                        sort_order=pymongo.DESCENDING
-                    )
-                    if last_trip_ts != -1:
-                        user['last_trip'] = arrow.get(last_trip_ts).format(time_format)
-    
-                    last_call_ts = ts.get_first_value_for_field(
-                        key='stats/server_api_time',
-                        field='data.ts',
-                        sort_order=pymongo.DESCENDING
-                    )
-                    if last_call_ts != -1:
-                        user['last_call'] = arrow.get(last_call_ts).format(time_format)
+                # Retrieve and assign pipeline range
+                pipeline_range = profile_data.get('pipeline_range', {})
+                start_ts = pipeline_range.get('start_ts')
+                end_ts = pipeline_range.get('end_ts')
+                if start_ts:
+                    user['first_trip'] = arrow.get(start_ts).format(time_format)
+                if end_ts:
+                    user['last_trip'] = arrow.get(end_ts).format(time_format)
+                
+                # Retrieve and assign last API call timestamp
+                last_call_ts = profile_data.get('last_call_ts')
+                if last_call_ts:
+                    user['last_call'] = arrow.get(last_call_ts).format('YYYY-MM-DD')
                 
             esdsq.store_dashboard_time(
                 "admin/db_utils/add_user_stats/process_user",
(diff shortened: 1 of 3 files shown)


TeachMeTW:
@JGreenlee Please review

utils/db_utils.py
```diff
@@ -437,58 +437,51 @@ def add_user_stats(user_data, batch_size=5):
...
+                    # Optional: Format the last_call timestamp for readability
+                    if user['last_call']:
+                        user['formatted_last_call'] = arrow.get(user['last_call']).format('YYYY-MM-DD HH:mm:ss')
+                    else:
+                        user['formatted_last_call'] = None
+                else:
+                    # Handle cases where profile_data is not found
+                    logging.warning(f"No profile data found for user_uuid: {user_uuid}")
+                    
+                    # Assign default values to ensure the user dictionary remains consistent
```
JGreenlee, line 472: I noticed that you already provided default values in the code above using `profile_data.get()` to handle cases where fields are missing from `profile_data`.

Then, you have this else block to handle when `profile_data` is None, and you give the same default values.

I think you can simpify this by setting `profile_data` to `{}` (an empty dict) if it is `None`. then you don't need the if-else block

This is better because i) it is more concise and you do not repeat the default values; ii) the code is flatter (less things nested in an if-else statement)

JGreenlee (commented):
Looks pretty good – I have a suggestion

utils/db_utils.py
TeachMeTW: @JGreenlee Addressed

JGreenlee:
I didn't do a great job of explaining what I was looking for since I wrote it out on my phone. Here's what I meant by my last suggestion:


```python
                # Fetch data for the user, cached for repeated queries
                if not profile_data:
                    profile_data = {}
                
                # Assign existing profile attributes to the user dictionary
                user['platform'] = profile_data.get('curr_platform')
                user['manufacturer'] = profile_data.get('manufacturer')
                user['app_version'] = profile_data.get('client_app_version')
                user['os_version'] = profile_data.get('client_os_version')
                user['phone_lang'] = profile_data.get('phone_lang')

                
                # Assign newly stored statistics to the user dictionary
                user['total_trips'] = profile_data.get('total_trips', 0)
                user['labeled_trips'] = profile_data.get('labeled_trips', 0)
                
                # Retrieve and assign pipeline range
                pipeline_range = profile_data.get('pipeline_range', {})
                user['pipeline_start_ts'] = pipeline_range.get('start_ts')
                user['pipeline_end_ts'] = pipeline_range.get('end_ts')
                
                # Retrieve and assign last API call timestamp
                user['last_call'] = profile_data.get('last_call')
                    
                # Optional: Format the last_call timestamp for readability
                if user['last_call']:
                    user['formatted_last_call'] = arrow.get(user['last_call']).format('YYYY-MM-DD HH:mm:ss')
                else:
                    user['formatted_last_call'] = None
                
```

If `profile_data` is falsy/None, it gets set to `{}` **before** we attempt to access its properties.

This allows the `.get()` calls to safely handle defaults, and we don't have to do all of this inside an `if` block.

TeachMeTW:
@JGreenlee  Addressed comment

JGreenlee (commented):
Similar to my review of https://github.com/e-mission/e-mission-server/pull/993, the code looks good but I am not sure it has been tested locally.

JGreenlee:
With this PR and https://github.com/e-mission/e-mission-server/pull/993, I ran the pipeline on one fresh test user.
I used `shankari_2016-07-22` and `shankari_2016-07-25` giving 6 trips.

image

There are blank fields here, which makes sense because I did not yet access the diary.

<hr>

I used the Simulator to run e-mission-phone and sign into this user. After refreshing the dashboard I see this:

image

<hr>

`last_call` is still blank. Also, there are several missing columns. Where are `pipeline_start_ts`, `pipeline_end_ts`, and `formatted_last_call` ?

JGreenlee:
`valid_uuids_cols` in `constants.py` controls what columns of the df get shown in the table:

https://github.com/e-mission/op-admin-dashboard/blob/d372913e80f771237dd2219819435a708f22f3e3/utils/constants.py#L38-L52

Either use the original column names or update the valid cols here.

(side note, maybe rename to `VALID_UUIDS_COLS` to match the other constants in `constants.py`)

JGreenlee:
After accessing the app via the Simulator, I ran the pipeline again trying to get `last_call` to populate. It was still blank.

<hr>

To debug I looked at pipeline logs to see if it was being recorded. Searching for `"last_call"` in `intake_single.log`, here's what I found:
```
2024-12-04 12:31:17,297:DEBUG:140704732739904:Last call timestamp: 1733333418.8644428
2024-12-04 12:31:17,297:INFO:140704732739904:Last call timestamp: 1733333418.8644428
2024-12-04 12:31:17,299:DEBUG:140704732739904:Updating user 4fea8afc-f4b1-4b3a-af09-d4bf7ef9e7e3 with fields {'pipeline_range': {'start_ts': 1469204768.0, 'end_ts': 1469493031.0}, 'total_trips': 6, 'labeled_trips': 0, 'last_call_ts': 1733333418.8644428}
2024-12-04 12:31:17,300:DEBUG:140704732739904:User profile updated with data: {'pipeline_range': {'start_ts': 1469204768.0, 'end_ts': 1469493031.0}, 'total_trips': 6, 'labeled_trips': 0, 'last_call_ts': 1733333418.8644428}
```

It is recorded as `last_call_ts` but the dashboard is looking for `last_call`

JGreenlee:
Note that I formatted `last_call` as YYYY-MM-DD (date only, no time). If you recall, we acknowledged that this value may lag behind by a bit because the pipeline run hourly, not updated in real time. We decided date-only is good enough because the use case here is just to determine if users are active or not.

Also note that I removed some default values, such as 0 for `total_trips` and `labeled_trips`. Think about it – there is a difference between i) the pipeline having run and recorded 0 labeled trips in the user stats and ii) the pipeline has not run and has not recorded any value.

In scenario (i) we'd want to show the 0 value but in scenario (ii) I'd argue that a blank cell is a better choice, which is what happens if we leave it `None`

TeachMeTW:
Tested via 

```./setup/activate.sh
./e-mission-py.bash bin/debug/load_timeline_for_day_and_user.py emission/tests/data/real_examples/shankari_2016-07-25 nrelop_dev-emulator-study_2
./e-mission-py.bash bin/debug/intake_single_user.py -e nrelop_dev-emulator-study_2
 
./e-mission-py.bash bin/debug/purge_user.py -e nrelop_dev-emulator-study_2```

On stm_ebike dump which had no uuids originally

TeachMeTW:
@JGreenlee  Ready for rereview

JGreenlee (approved):
LGTM but depends on https://github.com/e-mission/e-mission-server/pull/993 to be deployed at the same time

shankari:
@TeachMeTW the change is pretty straightforward, but the commit history is not.
I am going to rebase now, but please make sure to tidy up the commit history **in all repos**

Here's the first commit in the history here, for example
Screenshot 2024-12-21 at 5 15 56 PM

shankari:
The change was fairly complex to rebase because the first change commented out a bunch of code
https://github.com/e-mission/op-admin-dashboard/pull/153#issuecomment-2558289881
which was then reinstated in a different way in d241d375d833f93cd6de706a6bd8f5fbb32aab86
and the first change here was committed before d241d375d833f93cd6de706a6bd8f5fbb32aab86

Since the commits here were interspersed with commits that had already been merged, we needed to reorder before we could squash. But reordering wouldn't work because the base for the first commit here was the old timing, and so applying the change resulted in a merge conflict.

@TeachMeTW you really need to use have cleaner commits instead of pushing changes whenever you make any temporary change.

PR #154, opened 2024-12-13
improve batch loading of UUIDs stats

JGreenlee:
https://dash.plotly.com/background-callbacks#using-set_props-within-a-callback

Implements a 'background' callback to improve the batching of add_user_stats. This avoids relying on a dcc.Interval with a fixed time between batches

The new callback, which has background=True, continues running continuously on the server until all of the chunks have been processed. At the end of each chunk it calls set_props, which sends an update to the UI with the latest data for 'store-loaded-uuids'

This should speed things up further and prevent any glitches caused by the fixed interval

### DIFF
--- a/app_sidebar_collapsible.py
+++ b/app_sidebar_collapsible.py
@@ -14,7 +14,7 @@ import arrow
 
 import dash
 import dash_bootstrap_components as dbc
-from dash import Input, Output, dcc, html, Dash
+from dash import Input, Output, dcc, html, Dash, DiskcacheManager
 import dash_auth
 import logging
 import base64
@@ -42,10 +42,17 @@ elif auth_type == 'basic':
         'hello': 'world'
     }
 
+# diskcache manager; required for 'background' callbacks
+# https: // dash.plotly.com/background-callbacks
+import diskcache
+cache = diskcache.Cache("./cache")
+background_callback_manager = DiskcacheManager(cache)
+
 app = Dash(
     external_stylesheets=[dbc.themes.BOOTSTRAP, dbc.icons.FONT_AWESOME],
     suppress_callback_exceptions=True,
     use_pages=True,
+    background_callback_manager=background_callback_manager
 )
 server = app.server  # expose server variable for Procfile
 
--- a/pages/data.py
+++ b/pages/data.py
@@ -31,7 +31,6 @@ layout = html.Div(
         ]),
         html.Div(id='tabs-content'),
         dcc.Store(id='selected-tab', data='tab-uuids-datatable'),  # Store to hold selected tab
-        dcc.Store(id='store-uuids', data=[]),  # Store to hold the original UUIDs data
         dcc.Store(id='loaded-uuids-stats', data=[]),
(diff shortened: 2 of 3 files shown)


JGreenlee:
@TeachMeTW at your convenience, could you test this locally on the `stage` dataset and let me know how it works?

I tested it with smaller ones and it looks promising, but I am not able to easily load large dumps

TeachMeTW:
@JGreenlee

I've tested the application locally on the staging dataset and wanted to share my findings:

### Error on Overview Page

```A nonexistent object was used in an `Input` of a Dash callback. The id of this object is `tabs-datatable` and the property is `value`. The string ids in the current layout are: [url, store-trips, store-uuids, store-excluded-uuids, store-demographics, store-trajectories, page-content, date-picker, collapse-button, collapse-icon, collapse-filters, date-picker-timezone, excluded-subgroups, global-loading, _pages_location, _pages_content, _pages_store, _pages_dummy, card-users, card-active-users, card-trips, fig-sign-up-trend, fig-trips-trend]```


### Performance Observations

- **Staging Dataset (82 UUIDs):** *Lightning fast*
- **OpenAccess Dataset (200+ UUIDs):** *Normal/the usual performance*

It appears that the issue might be related to how larger datasets are being handled.

#### The superfast staging dataset
https://github.com/user-attachments/assets/14811ce9-b327-4ae6-bcbb-d75066eed28d

TeachMeTW:
@JGreenlee 

https://github.com/user-attachments/assets/ff322c41-0a58-4b2f-9ff0-3d520efc75b2

Tested on open_access. The UUIDs do not load in batches, they stay in a blank screen for 3-4 min then load everything at once -- akin to pre-batch behavior.

JGreenlee:
> @JGreenlee 
> 
> https://github.com/user-attachments/assets/ff322c41-0a58-4b2f-9ff0-3d520efc75b2
> 
> Tested on open_access. The UUIDs do not load in batches, they stay in a blank screen for 3-4 min then load everything at once -- akin to pre-batch behavior.

Thanks for testing. I'll check it out tomorrow

JGreenlee:
I think the reason for it being fast on Stage but slow on Laos + Open Access is simply that Stage has not nearly as many trips as Laos + Open Access.
I had been comparing the # of users, which Laos + Open Access have 2-3x more than Stage. However, the most expensive task here is querying for trips. Laos + Open Access both have like 10x more trips than Stage.

JGreenlee:
One hiccup I noticed that is annoying for the dev workflow is that autoreload sometimes breaks. Normally, when a file is modified and the dashboard is running, it refreshes itself. But if I do this while the UUIDs tab is showing, it gets stuck and I have to restart the the docker container to see my changes.

But if I move off of the UUIDs tab and then modify a file, it reloads like normal. So I think something about having a background callback running that blocks the autoreload process.

TeachMeTW:
@JGreenlee Tested with openaccess, it does batch loading as intended. It also resets progress as you said when switching between tabs

PR #157, opened 2025-01-06
🎛️ Roundup of UX improvements

JGreenlee:
Addresses backlog of UX issues in https://github.com/e-mission/e-mission-docs/issues/1066

#### show user_token on tables on data page dc1d590972f2aeb22a43cb3245df053c0e6cbd2e

> The UUIDs table gives a mapping of uuids to tokens, but the other tables only have uuids
> Instead of having to constanly cross-reference, which is tedious, we can just insert the user_token column alongside the user_id column in any table that has it

#### use dash-ag-grid for filterable tables f78069d34ebb84ad9abcddc89fb2b90806ce133b

> Since this dashboard was created, the dash team has released dash-ag-grid (https://dash.plotly.com/dash-ag-grid) which has many advantages over DataTable including better performance, more customiziability, and advanced sorting and filtering
> I was mainly interested just looking for a polished filtering mechanism, but I think this will serve us well in many ways; we may consider changing the DataTables on other pages to also be AgGrids

#### fix mode filters on maps 82d8064126e116d2621bfc13e52e09f442671367

> `data.user_input_mode_confirm` does not exist in the trips store, only `mode_confirm`. It was probably changed at some point and forgotten to change here

#### overhaul filters on map page 6295d1a8e93e19c3550a96aa27a333ca82b96c8e

> Simplify user filter dropdowns; unify "User UUIDs" and "User Emails" into one dropdown "Users". Still respects separate permissions for uuid / email; if options_emails permission is False, the admin will only see the UUIDs, and vice versa
> Made this text monospace to avoid ambiguous character in UUIDs/tokens
> 
> Mode filters use rich mode colors and icons using emcommon and label-options config
> New package dash-iconify allows use of MDI icon library (which is what rich modes use)

#### auto-adjust zoom on map 8063f537519892994579852247cc03e63d410786

> When the map was updated, it would just zoom in on whatever was the first location point, which makes for an annoying UX if you are trying to look at a lot of data spread over a large area.
> 
> These changes set the view of the map to the bounding box of the points being shown.
> This isn't supported natively by dash so there's a new function, thanks to folks on plotly forum

#### use eacd.get_dynamic_config() for permissions.py 512f9e43baa173f98fa5d00b88254f4a4728ca42

> no reason to repeat this code when e-mission-server has a function for that

#### download label-options config in background callback 33c68772fa2b9727085f2351959f321dadbb0800

> The label-options config will be used on the Map page to get colors and icons for rich modes. Later, we might also need it for footprint calculations if we add those to admin dash later
> 
> I made it a background callback so it does not slow the initialization of the dashboard


JGreenlee:
@TeachMeTW 
https://github.com/e-mission/e-mission-docs/issues/1066#issuecomment-2108185831

utils/ux_utils.py
```diff
@@ -0,0 +1,10 @@
+# ux_utils.py
+import dash_mantine_components as dmc
+
+def wrap_with_skeleton(component_id, height, children_component):
+    return dmc.Skeleton(
+        height=height,
+        visible=True,  # Initially visible
+        id=f'skeleton-{component_id}',
```
JGreenlee, line 8: are the skeleton components required to have an `id`?

pages/home.py
```diff
@@ -34,23 +35,35 @@
...
-        ])
-    ]
-)
+        [
+            dcc.Markdown(intro),
+            html.Div(id='home-page-load', children='', style={'display': 'none'}),
+            # Cards Section
+            dbc.Row([
+                dbc.Col(
+                    wrap_with_skeleton('users', 100, html.Div(id='card-users'))
```
JGreenlee, line 44: Although `wrap_with_skeleton` is nice and descriptive, could we consider a shorter name like `lazy`?
Because this function will repeatedly be used inline in the "markup" that defines the component tree, I am worried about the effect on readability

requirements.txt
```diff
@@ -16,3 +16,12 @@ flask-talisman==1.0.0
...
 dash-leaflet==1.0.7
+# dash[diskcache] dependencies
+# from https://github.com/plotly/dash/blob/dev/requirements/diskcache.txt
+diskcache>=5.2.1
+multiprocess>=0.70.12
+psutil>=5.8.0
+dash-ag-grid>=31.3.0
+dash-iconify>=0.1.2
+dash-mantine-components>=0.15.1
+dash[diskcache]
```
JGreenlee, line 27: Can you explain why `dash[diskcache]` was added?

I tried doing this before and it didn't seem to work, which is why I manually added `diskcache`, `multiprocess`, and `psutil` above as a workaround, which I believe accomplishes the same thing

Whichever way we go, it seems like we should only need one or the other

utils/ux_utils.py
TeachMeTW: Looking at documentation I believe not -- maybe thats why there were callback issues and bloat; I will remove and see if it improves things.

app_sidebar_collapsible.py
```diff
@@ -42,10 +47,17 @@
         'hello': 'world'
     }
 
+# diskcache manager; required for 'background' callbacks
+# https: // dash.plotly.com/background-callbacks
+import diskcache
+cache = diskcache.Cache("./cache")
+background_callback_manager = DiskcacheManager(cache)
+
 app = Dash(
```
TeachMeTW, line 56: @JGreenlee it was for this section; it was giving me an error and asked for diskcache

utils/ux_utils.py
TeachMeTW: Wait no, I remembered why i needed them; for the home page I am updating their visibility to have uniform loading. The callback needs the id hence why its needed

TeachMeTW: Now it is quite bloaty, but we could either forgo uniform loading and have it async wrt each other or keep as is. What are your thoughts @JGreenlee

pages/home.py
TeachMeTW: @JGreenlee wouldn't lazy be non-accurate description? I believe `skeleton` would suffice

utils/ux_utils.py
TeachMeTW: @JGreenlee Refactored it on latest commit; will rebase now to cut down on churn

pages/home.py
```diff
@@ -5,7 +5,7 @@
 
 """
 from uuid import UUID
-from dash import dcc, html, Input, Output, callback, register_page
+from dash import dcc, html, Input, Output, callback, register_page, no_update
```
JGreenlee, line 8: unused import: I removed it in 616f7da2794b71d2eeeb5530c5001db58df1527c and it was added back in 885f2febbfec0db59c5a68b267d11c91f5199af3 for seemingly no reason?

pages/data.py
```diff
@@ -156,13 +185,17 @@ def update_uuids_page_current(page_current, selected_tab):
...
+    Input('loaded-uuids-stats', 'data'),
 )
-def render_content(tab, store_uuids, store_excluded_uuids, store_trips, store_demographics, store_trajectories, start_date, end_date, timezone, n_intervals, key_list, current_page, loaded_uuids_store, all_data_loaded):
+def render_content(tab, store_uuids, store_excluded_uuids, store_trips, store_demographics, store_trajectories, start_date, end_date, timezone, key_list, current_page, loaded_uuids):
+    # Check for None values
+    inputs = [tab, store_uuids, store_excluded_uuids, store_trips, store_demographics, store_trajectories,
+              start_date, end_date, timezone, key_list, current_page, loaded_uuids]
+    if any(val is None for val in inputs):
+        raise PreventUpdate
+
```
JGreenlee, line 198: Why is this needed? I didn't see any explanation in the commit

pages/data.py
```diff
@@ -360,24 +363,27 @@ def render_content(tab, store_uuids, store_excluded_uuids, store_trips, store_de
...
     Input('subtabs-demographics', 'value'),
     Input('store-demographics', 'data'),
+    Input('store-uuids', 'data')
 )
-def update_sub_tab(tab, store_demographics):
+def update_sub_tab(tab, store_demographics, store_uuids):
+    # Check for None values
+    if tab is None or store_demographics is None or store_uuids is None:
+        raise PreventUpdate
+
```
JGreenlee, line 386: also here, same as above

pages/data.py
```diff
@@ -442,6 +448,9 @@ def update_sub_tab(tab, store_demographics):
 )
 # Controls visibility of columns in trips table and updates the label of button based on the number of clicks.
 def update_dropdowns_trips(n_clicks, button_label):
+    if n_clicks is None or button_label is None:
+        raise PreventUpdate
```
JGreenlee, line 452: and here

pages/data.py
```diff
@@ -479,29 +487,56 @@ def populate_datatable(df, table_id='', page_current=0):
...
+
+                    logging.info(f"Mapping df['user_id'] to uuids_df['user_id'] for user_token...")
+                    df['user_token'] = df['user_id'].map(uuids_df.set_index('user_id')['user_token']).fillna('Unknown')
+                else:
+                    logging.error(f"Neither 'data.user_id' nor 'user_id' columns exist in the DataFrame.")
+                    raise KeyError("Required columns for mapping 'user_token' are missing.")
 
+                logging.info(f"DataFrame after adding user_token: {df.head()}")
+            except KeyError as e:
+                logging.error(f"KeyError encountered: {e}")
```
JGreenlee, line 522: when is `user_token` not in the table? Is that expected or not?

JGreenlee:
I noticed a few regressions in this PR since I left off:

"Sign-ups trend" and "Trips trend" are now side-by-side. The full size charts were better because more fine-grained data could be distinguished on the x-axis

image

<hr>

The column names have the first letter capitalized (which does not match the underlying data), text is no longer in monospace (making opcodes/UUIDs ambiguous to read), and the text in the cells can no longer be copied
image

In 0eb122604874dfb68b9abfd1122545dcd7e805ac, I deliberately set the props of the table a certain way, but it was overwritten in 0eb122604874dfb68b9abfd1122545dcd7e805ac

pages/data.py
TeachMeTW: @JGreenlee It’s simply a guard that says “Wait until all required info is available before producing output.”; This avoids broken layouts or having to manually handle “loading states” within each callback

pages/home.py
TeachMeTW: @JGreenlee We need no update for `Update_All_Components` callback (skeleton)

pages/data.py
TeachMeTW: @JGreenlee There was a 'bug' where when switching between tabs, the df is not updated fast enough (my hunch) leading to a dash key error error. I believe user token is not in one of the other tabs like demographics or something else -- the bug entails switching between some tab and back to uuid tab

pages/data.py
TeachMeTW: ditto, sometimes when refreshing pages or switching tabs there is a chance you get a dash error. This prevents that

TeachMeTW:
@JGreenlee From the master/current, we have 

```
        dbc.Row([
            dcc.Graph(id="fig-sign-up-trend"),
            dcc.Graph(id="fig-trips-trend"),
        ])
 ```
  
  That I left unchanged, however, for some reason it has become next to each other as you state above.  Maybe its due to skeleton, maybe its due to mantine or something else. HOWEVER, I modified it so:
  
  ```
  dbc.Row(
    dbc.Col(html.Div(id="fig-sign-up-trend", children=skeleton(300))),
    className="mb-4"
),
dbc.Row(
    dbc.Col(html.Div(id="fig-trips-trend", children=skeleton(300)))
),
```

This will ensure -- i dare say guarantees -- the charts are stacked vertically instead of sitting next to each other

TeachMeTW:
@JGreenlee Please turn into ready for review please at earliest convenience -- I do not have perms to switch

JGreenlee:
Moved out of draft mode – but I still want to test locally before sending it over for @ shankari to review

JGreenlee:
I tested this branch and noticed one last issue, which is that the "raw/humanized units" toggle on the trips table no longer worked since changing to Dash Ag Grid.
I addressed this in 96cf080f3360340e3ae51eab2cc3066d669aff95

Everything appears to be working now:

image
image
image

pages/data.py
```diff
@@ -259,22 +229,28 @@ def render_content(tab, store_uuids, store_excluded_uuids, store_trips, store_de
...
+                    content = html.Div(
+                        [
+                            html.Div("No data available", style={'text-align': 'center', 'margin-bottom': '16px'}),
+                        ],
+                        style={'margin-top': '36px'}
+                    )
+
+                elif not has_perm:
+                    logging.debug(f"Callback - {selected_tab} Error Stage: No permission or no data available.")
+                    content = html.Div([html.P("No data available or you don't have permission.")])
```
shankari, line 243: The other two loads populate a `skeleton` if the dataframe is empty; why doesn't this need that code as well?

shankari (changes requested):
I am prioritizing this PR since it appears to be a blocker for https://github.com/e-mission/op-admin-dashboard/pull/164; and I really want to make sure that we show the new stats as the new deployments come on board.

JGreenlee:
I don't think this PR is blocking #164.

This PR depends on #154 and includes those commits. We should make a decision on whether batching the UUIDs table is needed anymore; if not then close #154 and update this PR accordingly

app_sidebar_collapsible.py
```diff
@@ -240,15 +256,22 @@ def make_home_page(): return [
...
-    dcc.Store(id='store-excluded-uuids', data={}), # list of UUIDs from excluded subgroups
-    dcc.Store(id='store-demographics', data={}),
-    dcc.Store(id='store-trajectories', data={}),
-    html.Div(id='page-content', children=make_home_page()),
-])
+def make_layout():
+    return dmc.MantineProvider(
+        theme={'colorScheme': 'light'},  # Optional: Customize theme
+        children=[
+            html.Div([
```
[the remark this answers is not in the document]
TeachMeTW: Without wrapping in a MantineProvider, components that rely on this context throw runtime errors and the components do not render at all

pages/data.py
TeachMeTW: I intentionally avoided using a skeleton for the slower-loading tabs because a skeleton might persist indefinitely, misleading users into thinking data is still loading. Instead, I opted for a more direct approach that either shows an empty state or the actual data, which better reflects the performance characteristics of those loads. The reason UUIDs tab was skeletonized because it had an updating callback (the batching), and the demographics tab was quite fast. The rest like trajectories or trips werent as fast and risked permanent skeleton even if the data was there.

JGreenlee:
I am removing the batch loading from the UUIDs table because:
1. We do not need it. All the information we need now is preloaded in the user profiles and loading them is very fast, even for hundreds of users
2. Using background callbacks requires a more complicated setup on production if we follow recommendations https://dash.plotly.com/background-callbacks

I am going to leave some bits of code, commented out, relating to background callbacks. This way, we can reconsider in the future if, for example, we run into issues with a program with thousands of users rather than just hundreds.

JGreenlee:
We read the user profiles for the Overview page but do not cache them. Then on the Data page we also read the user profiles to get the stats for the UUIDs table.
Since the Overview represents the entry point of the dashboard, we are always going to be loading the profiles upon initialization, so we might as well store that data to be used on the UUIDs table.

I think we should just read both the UUIDs DB and Profiles DB upfront, join them, and put them in one "users" store that can be used throughout the dashboard. This will be both simpler and more efficient.

app_sidebar_collapsible.py
```diff
@@ -339,6 +362,20 @@ def update_store_trips(start_date, end_date, timezone, excluded_uuids):
...
 
 
+@app.callback(
+    Output('store-label-options', 'data'),
+    Input('store-label-options', 'data'),
+    background=True,
+    cancel=[Input("_pages_location", "pathname")],
+)
+def load_label_options(_):
+    config = eacd.get_dynamic_config()
```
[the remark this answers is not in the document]
JGreenlee: Yeah I think moving `emission.analysis.configs.dynamic_config` to `emission.core.dynamic_config` makes sense

Then `ecd.get_dynamic_config()` can be used in the webapp, pipeline, and here

pages/data.py
```diff
@@ -133,20 +135,32 @@ def show_keylist_switch(tab):
...
+def load_uuids_stats(tab, uuids):
+    # slice uuids into chunks of 10
+    uuids_chunks = [uuids['data'][i:i+10] for i in range(0, len(uuids['data']), 10)]
+    loaded_stats = []
+    for uuids in uuids_chunks:
+        processed_uuids = db_utils.add_user_stats(uuids, 10)
+        loaded_stats.extend(processed_uuids)
+        logging.debug("loaded %s uuids stats: %s" % (len(loaded_stats), loaded_stats))
+        set_props('loaded-uuids-stats', {'data': loaded_stats})
+    return True
```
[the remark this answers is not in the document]
JGreenlee: Decided no batching for now; left comments in the code if we need it later

app_sidebar_collapsible.py
```diff
@@ -42,10 +50,18 @@
         'hello': 'world'
     }
 
+# diskcache manager; required for 'background' callbacks
+# https: // dash.plotly.com/background-callbacks
+import diskcache
+
+cache = diskcache.Cache("./cache")
+background_callback_manager = DiskcacheLongCallbackManager(cache)
```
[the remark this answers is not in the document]
JGreenlee: I misunderstood the docs at https://dash.plotly.com/background-callbacks and thought that the recommendation (Celery + Redis) was only available for Dash Enterprise. It seems that we could set up Celery + Redis manually if we wanted

However, given that we don't really need it anymore, I removed the background callbacks anyway

JGreenlee:
@TeachMeTW Can you include the new columns from #164 into this PR? It should just be a matter of adding them to constants.py, and if necessary formatting them as datetimes

If not I will get to it tomorrow

TeachMeTW:
So I've been trying to debug the profile df rn for a bit now
 
There is something wrong.

For example:
```
openpath_prod_dfc_fermata> db.Stage_Profiles.find({})
[
  {
    _id: ObjectId('67c7a8104af61e2171499a20'),
    user_id: Binary.createFromBase64('XN8pqUGwTN6hcALNIM32YQ==', 3),
    source: 'Shankari',
    create_ts: ISODate('2025-03-04T17:25:36.534Z'),
    update_ts: ISODate('2025-03-04T17:25:36.534Z'),
    mpg_array: [ 32.044384997811726 ],
    mode: {},
    purpose: {},
    labeled_trips: 0,
    pipeline_range: { start_ts: 1469618232.767, end_ts: 1469684686.181 },
    total_trips: 22
  }
]
 ```

last_call_ts does not show up anymore but it SHOULD.

According to the logs:
```
2025-03-04T17:26:10.284954-08:00**********UUID 5cdf29a9-41b0-4cde-a170-02cd20cdf661: creating composite objects **********
2025-03-04T17:26:11.033665-08:00**********UUID 5cdf29a9-41b0-4cde-a170-02cd20cdf661: storing user stats **********
INFO:root:User stats for 1fe45013-0fc3-4592-b883-214b5032e984: {
  "_id": "67c7a80e7d58a081b76185e0",
  "user_id": "1fe45013-0fc3-4592-b883-214b5032e984",
  "source": "Shankari",
  "create_ts": "2025-03-04 17:25:34.968000",
  "update_ts": "2025-03-04 17:25:34.968000",
  "mpg_array": [
    32.044384997811726
  ],
  "mode": {},
  "purpose": {}
}
INFO:root:Simulating API call for path: /TEST_PATH
DEBUG:root:START GET /TEST_PATH
DEBUG:root:update_last_call_timestamp called with: user_id=UUID('1fe45013-0fc3-4592-b883-214b5032e984'), call_path='/TEST_PATH'
DEBUG:root:Updating user 1fe45013-0fc3-4592-b883-214b5032e984 with fields {'last_call_ts': 1741137971.432677}
DEBUG:root:User profile updated with data: {'last_call_ts': 1741137971.432677}
DEBUG:root:New profile: {'_id': ObjectId('67c7a80e7d58a081b76185e0'), 'user_id': UUID('1fe45013-0fc3-4592-b883-214b5032e984'), 'source': 'Shankari', 'create_ts': datetime.datetime(2025, 3, 4, 17, 25, 34, 968000), 'update_ts': datetime.datetime(2025, 3, 4, 17, 25, 34, 968000), 'mpg_array': [32.044384997811726], 'mode': {}, 'purpose': {}, 'last_call_ts': 1741137971.432677}
DEBUG:root:END GET /TEST_PATH 1fe45013-0fc3-4592-b883-214b5032e984 0.00036215782165527344 
DEBUG:root:insert called with entry of type <class 'emission.core.wrapper.entry.Entry'>
DEBUG:root:entry was fine, no need to fix it
DEBUG:root:Inserting entry Entry({'_id': ObjectId('67c7a8337d58a081b76185e1'), 'user_id': UUID('1fe45013-0fc3-4592-b883-214b5032e984'), 'metadata': Metadata({'key': 'stats/server_api_time', 'platform': 'server', 'write_ts': 1741137971.437344, 'time_zone': 'America/Los_Angeles', 'write_local_dt': LocalDate({'year': 2025, 'month': 3, 'day': 4, 'hour': 17, 'minute': 26, 'second': 11, 'weekday': 1, 'timezone': 'America/Los_Angeles'}), 'write_fmt_time': '2025-03-04T17:26:11.437344-08:00'}), 'data': {'name': 'GET_/TEST_PATH', 'ts': 1741137971.432202, 'reading': 0.00036215782165527344}}) into timeseries
DEBUG:root:insert called with entry of type <class 'emission.core.wrapper.entry.Entry'>
DEBUG:root:entry was fine, no need to fix it
DEBUG:root:Inserting entry Entry({'_id': ObjectId('67c7a8337d58a081b76185e2'), 'user_id': UUID('1fe45013-0fc3-4592-b883-214b5032e984'), 'metadata': Metadata({'key': 'stats/server_api_time', 'platform': 'server', 'write_ts': 1741137971.440483, 'time_zone': 'America/Los_Angeles', 'write_local_dt': LocalDate({'year': 2025, 'month': 3, 'day': 4, 'hour': 17, 'minute': 26, 'second': 11, 'weekday': 1, 'timezone': 'America/Los_Angeles'}), 'write_fmt_time': '2025-03-04T17:26:11.440483-08:00'}), 'data': {'name': 'GET_/TEST_PATH_cputime', 'ts': 1741137971.432202, 'reading': 0.00035787499999173633}}) into timeseries
...
  "last_diary_fetch_ts": 1741137971.461808,
  "last_location_ts": 1500,
  "last_phone_data_ts": 2500
}
```

Especially according to:
```
DEBUG:root:New profile: {'_id': ObjectId('67c7a80e7d58a081b76185e0'), 'user_id': UUID('1fe45013-0fc3-4592-b883-214b5032e984'), 'source': 'Shankari', 'create_ts': datetime.datetime(2025, 3, 4, 17, 25, 34, 968000), 'update_ts': datetime.datetime(2025, 3, 4, 17, 25, 34, 968000), 'mpg_array': [32.044384997811726], 'mode': {}, 'purpose': {}, 'last_call_ts': 1741137971.432677}
```

Which is in user_stat.py:
```
def update_user_profile(user_id: str, data: Dict[str, Any]) -> None:
    """
    Updates the user profile with the provided data.

    :param user_id: The UUID of the user.
    :type user_id: str
    :param data: The data to update in the user profile.
    :type data: Dict[str, Any]
    :return: None
    """
    user = ecwu.User.fromUUID(user_id)
    user.update(data)
    logging.debug(f"User profile updated with data: {data}")
    logging.debug(f"New profile: {user.getProfile()}")
 ```

Yet when I go to the DB, that 'new' profile is NOT there.

But when running `./e-mission-py.bash -m unittest -v emission.tests.analysisTests.intakeTests.TestUserStat.TestUserStats.testLastCall` it works as intended.

TeachMeTW:
I am using the same tests I did in https://github.com/e-mission/op-admin-dashboard/pull/164#issuecomment-2666996429 which SHOULD work but it no longer does. Debugging now.

TeachMeTW:
According to the logs it should be inserting, but it is not:

```
2025-03-04T17:54:58.211505-08:00**********UUID 1bffdb93-87a7-4e8f-b670-c7a46326329d: segmenting into sections**********
2025-03-04T17:54:58.228798-08:00**********UUID 1bffdb93-87a7-4e8f-b670-c7a46326329d: smoothing sections**********
2025-03-04T17:54:58.244179-08:00**********UUID 1bffdb93-87a7-4e8f-b670-c7a46326329d: cleaning and resampling timeline**********
2025-03-04T17:54:58.268745-08:00**********UUID 1bffdb93-87a7-4e8f-b670-c7a46326329d: inferring transportation mode**********
2025-03-04T17:54:58.284105-08:00**********UUID 1bffdb93-87a7-4e8f-b670-c7a46326329d: inferring labels**********
2025-03-04T17:54:58.299074-08:00**********UUID 1bffdb93-87a7-4e8f-b670-c7a46326329d: populating expectations**********
2025-03-04T17:54:58.314781-08:00**********UUID 1bffdb93-87a7-4e8f-b670-c7a46326329d: creating confirmed objects **********
2025-03-04T17:54:58.337540-08:00**********UUID 1bffdb93-87a7-4e8f-b670-c7a46326329d: creating composite objects **********
2025-03-04T17:54:58.407401-08:00**********UUID 1bffdb93-87a7-4e8f-b670-c7a46326329d: storing user stats **********
DEBUG:root:finished querying values for ['analysis/confirmed_trip'], count = 0
DEBUG:root:finished querying values for [], count = 0
DEBUG:root:finished querying values for ['analysis/confirmed_trip'], count = 0
INFO:root:Total trips: 0, Labeled trips: 0
INFO:root:user_id type: <class 'uuid.UUID'>
DEBUG:root:Updating user 1bffdb93-87a7-4e8f-b670-c7a46326329d with fields {'pipeline_range': {'start_ts': None, 'end_ts': None}, 'total_trips': 0, 'labeled_trips': 0}
DEBUG:root:User profile updated with data: {'pipeline_range': {'start_ts': None, 'end_ts': None}, 'total_trips': 0, 'labeled_trips': 0}
DEBUG:root:New profile: {'_id': ObjectId('67c7aef17d58a081b7618636'), 'user_id': UUID('1bffdb93-87a7-4e8f-b670-c7a46326329d'), 'source': 'Shankari', 'create_ts': datetime.datetime(2025, 3, 4, 17, 54, 57, 992000), 'update_ts': datetime.datetime(2025, 3, 4, 17, 54, 57, 992000), 'mpg_array': [32.044384997811726], 'mode': {}, 'purpose': {}, 'last_call_ts': 1741139698.058085, 'last_sync_ts': 1741139698.049193, 'last_put_ts': 1741139698.049193, 'last_diary_fetch_ts': 1741139698.058085, 'last_location_ts': 1500, 'last_phone_data_ts': 2500, 'labeled_trips': 0, 'pipeline_range': {'start_ts': None, 'end_ts': None}, 'total_trips': 0}
DEBUG:root:User profile updated successfully.
DEBUG:root:insert called with entry of type <class 'emission.core.wrapper.entry.Entry'>
DEBUG:root:entry was fine, no need to fix it
DEBUG:root:Inserting entry Entry({'_id': ObjectId('67c7aef27d58a081b7618659'), 'user_id': UUID('1bffdb93-87a7-4e8f-b670-c7a46326329d'), 'metadata': Metadata({'key': 'stats/pipeline_time', 'platform': 'server', 'write_ts': 1741139698.420366, 'time_zone': 'America/Los_Angeles', 'write_local_dt': LocalDate({'year': 2025, 'month': 3, 'day': 4, 'hour': 17, 'minute': 54, 'second': 58, 'weekday': 1, 'timezone': 'America/Los_Angeles'}), 'write_fmt_time': '2025-03-04T17:54:58.420366-08:00'}), 'data': {'name': 'STORE_PIPELINE_DEPENDENT_USER_STATS', 'ts': 1741139698.4202921, 'reading': 0.013169834000109404}}) into timeseries
```



```
openpath_prod_dfc_fermata> db.Stage_Profiles.find({})

openpath_prod_dfc_fermata> 
```

TeachMeTW:
I believe I got it to work:
```
  {
    _id: ObjectId('67c7b19d6daf4bdd5246f718'),
    user_id: Binary.createFromBase64('cdRsHSXjS66BTRTn7/pY0Q==', 3),
    source: 'Shankari',
    create_ts: ISODate('2025-03-04T18:06:21.505Z'),
    update_ts: ISODate('2025-03-04T18:06:21.966Z'),
    mpg_array: [ 32.044384997811726 ],
    mode: {},
    purpose: {},
    labeled_trips: 0,
    pipeline_range: { start_ts: 1469618232.767, end_ts: 1469684686.181 },
    total_trips: 22,
    last_call_ts: 1741140425.986909,
    last_sync_ts: 1741140425.977124,
    last_put_ts: 1741140425.977124,
    last_diary_fetch_ts: 1741140425.986909,
    last_location_ts: 1500,
    last_phone_data_ts: 2500
  }
  ```
  
  
  It was due to ordering differences; aka run pipeline ought to go last or twice depending on configuration

File: pages/data.py [hunk 15]
```diff
@@ -441,11 +442,15 @@ def populate_datatable(df, store_uuids, table_id):
             # Ag Grid does not allow . in column names; replace with _
             # before creating the DataTable
             df.columns = [col.replace('.', '_') for col in df.columns]
+            column_defs = [
+                {"field": col, "headerName": col.replace("_", " ").title()}
+                for col in VALID_UUIDS_COLS
```
JGreenlee, line 447: This function (`populate_datatable`) is called for all the tables on the data page, not just UUIDs, so we should definitely not be filtering these columns according to `VALID_UUIDS_COLS`.

Also, why replace underscores with spaces?

utils/db_utils.py
```diff
@@ -55,71 +55,48 @@ def df_to_filtered_records(df, col_to_filter=None, vals_to_exclude: list[str] =
...
+        logging.debug("Querying for all User Profiles")
+        profiles_entries = edb.get_profile_db().find()
+        profiles_df = pd.json_normalize(list(profiles_entries))
+        profiles_df['user_id'] = profiles_df['user_id'].apply(str)
+        profiles_df['last_sync_ts'] = pd.to_datetime(profiles_df['last_sync_ts'])
+        profiles_df['last_location_ts'] = pd.to_datetime(profiles_df['last_location_ts'])
+        profiles_df['last_phone_data_ts'] = pd.to_datetime(profiles_df['last_phone_data_ts'])
+        profiles_df['last_diary_fetch_ts'] = pd.to_datetime(profiles_df['last_diary_fetch_ts'])
+        profiles_df['create_ts'] = pd.to_datetime(profiles_df['create_ts'])
+        profiles_df['update_ts'] = pd.to_datetime(profiles_df['update_ts'])
```
JGreenlee, line 83: When I tested this I got an error:
`ValueError: mixed datetimes and integers in passed array`

pages/data.py
TeachMeTW: Ah got it. I replaced the underscores with spaces to make it 'cleaner'/human readable

utils/db_utils.py
TeachMeTW: @JGreenlee Which dataset are you getting an error? Let me try to reproduce.

JGreenlee:
I will take care of this tonight

TeachMeTW:
@JGreenlee Try my latest commit first; I believe it should be good now?

TeachMeTW:
I tried it with open_access

shankari:
> It was due to ordering differences; aka run pipeline ought to go last or twice depending on configuration

I am not sure what you mean by this

> I am using the same tests I did in https://github.com/e-mission/op-admin-dashboard/pull/164#issuecomment-2666996429 which SHOULD work but it no longer does. Debugging now.

Given that we have merged and deployed https://github.com/e-mission/e-mission-server/pull/1024 **and** https://github.com/e-mission/e-mission-server/pull/1025 why do you need to use the same tests as before? Any recent snapshot will have the values already filled in

JGreenlee:
@TeachMeTW I saw your commits but I was working on a less complex solution. I added the new columns and fixed a few other things too.
I think this PR includes everything now.

<hr>

Testing done:

- Generated a new opcode

Only these columns show up:

image

(expected)

<hr>

- Generated a new opcode
- Loaded `shankari_2016-07-25`
- Ran the pipeline
- Signed in on the Simulator
- Labeled one trip
- Did a Force Sync
- Ran the pipeline again

Now all these columns show up:

image
image
image

JGreenlee:
I did a run-through on a dump of ccebikes, which I took on 2/25. There is a profile with `last_call_ts` of `null`
After my most recent commit we will not error even if there are null/missing `last_*_ts` fields

shankari:
> I also tested on a dump of ccebikes, which I took on 2/25. I noticed a profile with last_call_ts of null.
> After my most recent commit we will not error even if there are null/missing last_*_ts fields

Yes, there was one user that had two profiles for the same UUID, IIRC

pages/data.py
shankari: the UUID tab does not have a callback any more.
I don't fully understand what a skeleton does (it looks from the response here that it is some kind of load screen).

As a future cleanup, can one of you clarify what this functionality does and how you are using it, with examples?

pages/data.py
```diff
@@ -479,29 +428,43 @@ def populate_datatable(df, table_id='', page_current=0):
             "admin/data/populate_datatable/check_dataframe_type",
             stage1_timer
         )
-
+        if 'user_token' not in df.columns:
```
shankari, line 431: nit, future fix: `user_opcode`? token can be used in multiple contexts

shankari (approved):
I have a couple of cleanup fixes, but let's finally get this to staging and unblock everything else

utils/permissions.py
```diff
@@ -66,47 +60,41 @@ def get_all_named_trip_columns():
...
+                    if col['path'] not in columns])
+    columns.extend([col for col in get_required_columns()
+                    if col not in columns])
     logging.debug("get_all_trip_columns: after additional columns, curr set is %s" % columns)
     return columns
 
 
 def get_allowed_trip_columns():
-    columns = set(constants.VALID_TRIP_COLS)
+    columns = list(constants.VALID_TRIP_COLS)
```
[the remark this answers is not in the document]
JGreenlee: > In Python, sets are unique and unordered.
> Several of the functions in permissions.py used sets to track the columns that show up, which ensures no duplicates.
> However, it causes inconsistent ordering of the columns and is bad UX.
> Changing these to lists and only adding columns that aren't already added gives an ordered list without duplicates.

app_sidebar_collapsible.py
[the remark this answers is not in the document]
JGreenlee: We are using [Skeleton Loading](https://www.dash-mantine-components.com/components/skeleton) from [Dash Mantine Components](https://www.dash-mantine-components.com/), allowing us to more clearly convey to the user when components are being lazily-loaded

For example, if one of the cards on the Overview is taking a long time to load, we want to show the cards that are ready while the remaining card has a pulsing placeholder element until it is finished loading

This is less of an issue now that everything on the Overview is fast, but I think it can still be useful for other parts of the dashboard (now and in the future)

Using Dash Mantine Components requires the app to be wrapped in a MantineProvider https://www.dash-mantine-components.com/components/mantineprovider

PR #163, opened 2025-02-05
Refactor `get_number_of_active_users` 

TeachMeTW:
## Summary
Refactored `get_number_of_active_users` to reduce the number of database calls from **O(n)** to **O(1)** by utilizing a single query with `$in`. The previous implementation iterated over each user UUID and made separate calls to fetch profile data, leading to inefficient database access.

## Changes
- Consolidated multiple DB calls into a single query using `$in` for batch retrieval.

## Testing
- Verified correctness via `db.Stage_uuids.find().pretty()`.
- Selected test UUID(s) and updated `db.Stage_Profiles` to modify `last_call_ts`, ensuring flagged users were correctly identified as active.

See https://github.com/e-mission/op-admin-dashboard/pull/156/files/9566ffa2973718d902d8d11ec218c05d1b39ac57#r1938894371

### DIFF
--- a/pages/home.py
+++ b/pages/home.py
@@ -133,17 +133,18 @@ def compute_trips_trend(trips_df, date_col):
 
 def get_number_of_active_users(uuid_list, threshold):
     with ect.Timer() as total_timer:
-        number_of_active_users = 0
-        current_timestamp = arrow.utcnow().timestamp()
-        for npu in uuid_list:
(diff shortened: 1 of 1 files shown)


JGreenlee:
This PR will not be needed because https://github.com/e-mission/op-admin-dashboard/pull/157 now reads all the profiles at once

PR #172, opened 2025-06-13
Chunked trajectory download

Tvpower:
@JGreenlee

### DIFF
--- a/utils/db_utils.py
+++ b/utils/db_utils.py
@@ -103,9 +103,6 @@ def query_confirmed_trips(start_date: str, end_date: str, tz: str):
             stage1_timer
         )
 
-        #explanation: when i take a 20 min walk with a # of minutes there is a big amount of data
-        #when a trip is labelled the data is saved here
-
         # Stage 2: Retrieve aggregate time series
         with ect.Timer() as stage2_timer:
             ts = esta.TimeSeries.get_aggregate_time_series()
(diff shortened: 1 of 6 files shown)


JGreenlee (changes requested):

utils/db_utils.py
```diff
@@ -486,8 +498,150 @@ def query_segments_crossing_endpoints(poly_region_start, poly_region_end, start_
...
 analysis_timeseries_db = edb.get_analysis_timeseries_db()
 
+
 # Fetches sensed_mode for each section in a list
 # sections format example: [{'section': ObjectId('648d02b227fd2bb6635414a0'), 'user_id': UUID('6d7edf29-8b3f-451b-8d66-984cb8dd8906')}]
 def query_inferred_sections_modes(sections):
     return esds.cleaned2inferred_section_list(sections)
 
+
+def query_entries_chunked(key_list, date_query,
```
JGreenlee, line 508: The current version of `query_entries_chunked` does not meet the requirements:

> This function will basically be a wrapper around `ts.find_entries` with the chunking logic around it. It should essentially take the same arguments as `find_entries`:
> - key_list
>   - for trajectories, that key was analysis/recreation_location or background/location. In your generic function, it could be anything, and that gets passed to `ts.find_entries`
> - time_query
> - geo_query
> - extra_query_list
> - the query limit (default = 250k)
> 
> return a list (or iterator) of all the combined entries (which may be > 250k)

Namely, it has `date_query` instead of `time_query`, and it yields and returns dataframes instead of simply returning a list.

PR #176, opened 2026-02-05
Updated data to display UserProfileSurvey data and parse through it's…

maggyparra:
In this PR I updated data to display the UserProfileSurvey survey data and parse through its XML file to display survey questions.

### DIFF
--- a/pages/data.py
+++ b/pages/data.py
@@ -205,7 +205,7 @@ def render_content(tab, store_uuids, store_excluded_uuids, store_trips, store_su
                 df = pd.DataFrame(data)
                 if df.empty and has_perm:
                     logging.debug(f"Callback - {selected_tab} loaded_trips is empty.")
-                    # Restoration of the correct empty state for the Trips tab
+                    
                     content = html.Div( 
                         [
                             html.Div("No data available", style={'text-align': 'center', 'margin-bottom': '16px'}), 
@@ -277,7 +277,8 @@ def render_content(tab, store_uuids, store_excluded_uuids, store_trips, store_su
                         content = skeleton(100)
                     else:
                         # Map the raw backend keys to the specific dfc-fermata survey names so they are fetched properly
-                        survey_label_map = {key: key.replace('Survey', '').replace('Dfc', 'dfc-').lower() for key in data.keys()}
+                        surveys_config = perm_utils.surveyinfo.get("surveys", {})
+                        survey_label_map = {key: surveys_config.get(key, {}).get("label", key) for key in data.keys()}
                         
                         content = html.Div([
                             dcc.Tabs(id='subtabs-surveys', value=list(data.keys())[0], children=[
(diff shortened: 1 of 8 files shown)


pages/data.py
```diff
@@ -50,79 +52,46 @@
...
-        total_timer
-    )
+                df['data.end_loc.coordinates'] = df['data.end_loc.coordinates'].apply(lambda x: f'({x[0]}, {x[1]})' if isinstance(x, list) else x)
+            esdsq.store_dashboard_time("admin/data/clean_location_data/clean_end_loc_coordinates", stage2_timer)
 
+    esdsq.store_dashboard_time("admin/db_utils/clean_location_data/total_time", total_timer)
     return df
 
-def update_store_trajectories(start_date: str, end_date: str, tz: str, excluded_uuids, key_list):
+def update_store_trajectories(start_date, end_date, tz, excluded_uuids, key_list):
```
JGreenlee: Please leave the type annotations; they are very helpful!

JGreenlee (changes requested):
Thanks for the PR!
In general I am noticing a lot of changes that seem unrelated to the new visualizations (like whitespace changes, reformats, removing comments and type annotations, etc.)

This is undesirable because i) A lot of the things that have been removed are useful and important, and ii) It makes it harder to review the main changes that this PR accomplishes because there is a lot of "noise" for me to sift through while reviewing

What you showed me on Monday looked great, so I know the functionality is there. but making a good PR is just as much about keeping it clean and tidy (i.e. only including the necessary changes) as it is about the functionality.

Let me come back to this after you have tidied up your changes. Good progress overall!

JGreenlee (changes requested):

maggyparra:
Screenshot 2026-02-16 at 4 31 02 PM
Screenshot 2026-02-16 at 4 31 33 PM

maggyparra:
Screenshot 2026-02-18 at 10 37 28 AM
Screenshot 2026-02-18 at 10 37 43 AM

JGreenlee:
The labels in the survey are:
- "Disagree"
- "-"
- "Neutral"
- "-"
- "Agree" 

And the values are numbers 1-5.

For the legend, instead of replacing the values with the labels (and ending up with "-" in the legend), it would be useful to represent it as:
- "1 (Disagree)"
- "2"
- "3 (Neutral)"
- "4"
- "5 (Agree)"

This way the user can see the text values but also the numerical values for 2 and 4, which don't have text values

JGreenlee (changes requested):
Commenting on the latest version of the code that has been pushed (though I think this is older than your recent screenshots)

JGreenlee (changes requested):
Once you address these places where the survey names have been hardcoded, we can move onto implementing similar visualizations on the public dashboard https://github.com/e-mission/em-public-dashboard

JGreenlee (changes requested):

PR #178, opened 2026-06-26
Convert this to multiplatform builds as well

shankari:
This is a follow-on to e-mission/e-mission-server#1097 To tie up loose ends around e-mission/e-mission-docs#1143 by making the dashboards also multiplatform

We use a pattern very similar to https://github.com/e-mission/em-public-dashboard/pull/204 because this repo also uses `docker compose` to build images.

The only difference is that the push step is also different depending on the `event_name`. This is because, although the push command uses the cache, it is technically a rebuild, and needs to have the correct variables.

I tested this by building (with a correct `SERVER_IMAGE_TAG`) and storing to a cache and then rebuilding again with an incorrect setting (`SERVER_IMAGE_TAG=INVALID).

### DIFF
--- a/.github/workflows/image_build_push.yml
+++ b/.github/workflows/image_build_push.yml
@@ -2,7 +2,7 @@ name: docker-image-push-admin
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, '**__image_build_**' ]
 
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


shankari:
Screenshot 2026-06-26 at 7 39 24 AM

